PR #13, opened 2015-09-01
add some post-install instructions

maxamillion:


pbabinca:
I've merged this on the top of master by mistake. So I've rebased it against develop and pushed there. It'll land to master with the rest of the commits after they are tested properly.

PR #24, opened 2016-03-08
quickfix for downloads always being .tar

breillyrh:
rebased against origin/develop.


twaugh:
Longer term it would be nice for it to be able to handle `.tar.xz` as well, but for now this looks good.

PR #26, opened 2016-03-09
Compression Fix with new OSBS API Call

breillyrh:
Should be pulled in after this gets merged - https://github.com/projectatomic/osbs-client/pull/360

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -286,12 +286,22 @@ class CreateContainerTask(BaseTaskHandler):
                                                   self.nfs_dest_dir(),
                                                   source_filename)
 
-    def _download_files(self, source_filename, target_filename=None):
-        if not target_filename:
-            if source_filename.endswith(".tar.gz"):
-                target_filename = "image.tar.gz"
-            else:
-                target_filename = "image.tar"
+    def _getOutputImageTemplate(self, arch, compressed_extension=None):
+
+        try:
+            if not compressed_extension:
+                compressed_extension = self.osbs().get_compression_extension()
+        except AttributeError:
+            # for compatibility with osbs without this API
+            pass
+
+        # OSBS returns None if no extension
+        if not compressed_extension:
+            compressed_extension = ''
+
+        return "image-%s.tar%s" % (arch, compressed_extension)
+
+    def _download_files(self, source_filename, target_filename='image.tar'):
         localpath = os.path.join(self.workdir, target_filename)
         remote_url = self._get_file_url(source_filename)
         koji.ensuredir(self.workdir)
@@ -543,10 +553,8 @@ class CreateContainerTask(BaseTaskHandler):
         files = []
         if response.is_succeeded() and response.get_tar_metadata_filename():
             sourceimage = response.get_tar_metadata_filename()
-            if sourceimage.endswith(".tar.gz"):
-                targetimage = "image-%s.tar.gz" % arch
-            else:
-                targetimage = "image-%s.tar" % arch
+            targetimage = self._getOutputImageTemplate(arch)
+
             files = self._download_files(sourceimage, targetimage)
 
         rpmlist = []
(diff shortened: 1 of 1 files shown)


pbabinca:
Please rebase this against devel HEAD

twaugh:
The osbs api commit looks correct, just needs rebasing.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -727,6 +729,11 @@ def _getOutputImageTemplate(self, **kwargs):
                 have_all_fields = False
         if have_all_fields:
             output_template = "%(name)s-%(version)s-%(release)s-%(architecture)s.tar" % kwargs
+
+        extension = self.osbs().get_compression_extension()
+        if extension:
+            output_template = output_template + extension
```
pbabinca: Use string formatting here instead as this can easily fail if we got unicode from osbs and try to append it to str.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -543,10 +543,12 @@ def handler(self, src, target_info, arch, output_template, scratch=False,
...
             sourceimage = response.get_tar_metadata_filename()
-            if sourceimage.endswith(".tar.gz"):
-                targetimage = "image-%s.tar.gz" % arch
-            else:
-                targetimage = "image-%s.tar" % arch
+            targetimage = "image-%s.tar" % arch
+
+            extension = self.osbs().get_compression_extension()
+            if extension:
+                targetimage = targetimage + extension
```
pbabinca: Use string formatting here instead as this can easily fail if we got unicode from osbs and try to append it to str.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -288,10 +288,10 @@ def _get_file_url(self, source_filename):
...
     def _download_files(self, source_filename, target_filename=None):
         if not target_filename:
-            if source_filename.endswith(".tar.gz"):
-                target_filename = "image.tar.gz"
-            else:
-                target_filename = "image.tar"
+            target_filename = "image.tar"
+            extension = self.osbs().get_compression_extension()
+            if extension:
+                target_filename = target_filename + extension
```
pbabinca: Use string formatting here instead as this can easily fail if we got unicode from osbs and try to append it to str.

pbabinca:
Looking at the original code again I believe plugin will fail if the new OSBS API isn't available. Make sure this doesn't happen:
- check if there is such method there and if it isn't there don't use it (preferred way to solve) 
- explicitly require version of osbs which has this API

twaugh:
Yes, using `try: ... except AttributeError: pass` to handle it would be good.

breillyrh:
I've made the suggested changes, though I had to create two identical helper functions as this code spans two different classes.

pbabinca:
I still don't like that code. I've looked deeper into the logic and found the change needs to be for two parts:

1. in `CreateContainerTask` where arch-specific filename is created (`"image-%s.tar" % arch`)
2. in `BuildContainerTask` where real unique output filename is constructed (`"%(name)s-%(version)s-%(release)s-%(architecture)s.tar" % kwargs`)

That third place - in `_download_files()` method isn't needed as that's just an default argument if `target_filename` isn't set. As part of refactoring I guess it will be clearer to change method definition to either either to: `def _download_files(self, source_filename, target_filename='image.tar')` or require both arguments: `def _download_files(self, source_filename, target_filename)`.

`CreateContainerTask` class could get method (maybe `_getOutputImageTemplate`?) which constructs the whole template with `arch` and optional compress extension. I'd like to have single construction of such template string there like this:
```
try:
    compressed_extension = self.osbs().get_compression_extension()
except AttributeError:
    # for compatibility with osbs without this API
    compressed_extension = ''

return "image-%s.tar%s" % (arch, compressed_extension)
```

And `_getOutputImageTemplate` method in `BuildContainerTask` class to something like:
```
# ...
if have_all_fields:
    base_name "%(name)s-%(version)s-%(release)s-%(architecture)s" % kwargs
else:
    base_name = 'image'
# ... (get that compressed extension from osbs somehow)
return "%s.tar%s" % (base_name, compressed_extension)
```

And finally I don't need that commits as separate attempts to solve this issue so squash them to single commit where applicable.

breillyrh:
Changes made and commits merged.

PR #31, opened 2016-04-01
Rebase cg

maxamillion:
I did an interactive rebase and squashed everything down but for some reason there's still some left over commits showing here ...

PR #32, opened 2016-04-04
Rebase cg

maxamillion:
Carry over of https://github.com/release-engineering/koji-containerbuild/pull/31

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -117,64 +117,6 @@ class My_SCM(SCM):
         return git_uri
 
 
-class FileWatcher(object):
-    """Watch directory for new or changed files which can be iterated on
-
-    Rewritten mock() from Buildroot class of kojid. When modifying keep that in
-    mind and after the API looks stable enough try to merge the code back to
-    koji.
-    """
-    def __init__(self, result_dir, logger):
-        self._result_dir = result_dir
-        self.logger = logger
-        self._logs = {}
-
-    def _list_files(self):
-        try:
-            results = os.listdir(self._result_dir)
-        except OSError:
-            # will happen when mock hasn't created the resultdir yet
-            return
-
-        for fname in results:
-            if fname.endswith('.log') and fname not in self._logs:
-                fpath = os.path.join(self._result_dir, fname)
-                self._logs[fname] = (None, None, 0, fpath)
-
-    def _reopen_file(self, fname, fd, inode, size, fpath):
-        try:
-            stat_info = os.stat(fpath)
-            if not fd or stat_info.st_ino != inode or stat_info.st_size < size:
-                # either a file we haven't opened before, or mock replaced a file we had open with
-                # a new file and is writing to it, or truncated the file we're reading,
-                # but our fd is pointing to the previous location in the old file
-                if fd:
-                    self.logger.info('Rereading %s, inode: %s -> %s, size: %s -> %s' %
-                                     (fpath, inode, stat_info.st_ino, size, stat_info.st_size))
-                    fd.close()
-                fd = file(fpath, 'r')
-            self._logs[fname] = (fd, stat_info.st_ino, stat_info.st_size, fpath)
-        except:
-            self.logger.error("Error reading mock log: %s", fpath)
-            self.logger.error(''.join(traceback.format_exception(*sys.exc_info())))
-            return False
-        return fd
-
-    def files_to_upload(self):
-        self._list_files()
-
-        for (fname, (fd, inode, size, fpath)) in self._logs.items():
-            fd = self._reopen_file(fname, fd, inode, size, fpath)
-            if fd is False:
-                return
-            yield (fd, fname)
-
-    def clean(self):
-        for (fname, (fd, inode, size, fpath)) in self._logs.items():
-            if fd:
-                fd.close()
-
-
 class LabelsWrapper(object):
     def __init__(self, dockerfile_path, logger_name=None):
         self.dockerfile_path = dockerfile_path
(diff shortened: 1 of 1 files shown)


pbabinca:
This seems to only remove code and not add any more as in #31.

maxamillion:
@pbabinca The only addition of code was from Dennis, which was already merged into this branch so when I rebased and opened the new pull request it wasn't carried over again. All I'm doing is removing code that was importing content and/or handling scratch builds because the bulk of the work is now happening in the Content Generator import exit_koji_promote plugin in atomic-reactor. This plugin is basically just going to create the task and sub-task(s).

pbabinca:
Shouldn't tarball-related code removed too then?

maxamillion:
@pbabinca it can be, but it doesn't really effect anything since with v2 builds there are no tarballs. I didn't know if y'all had any reason to keep it for v1 so I didn't want to remove it.

pbabinca:
@maxamillion what's the point of 7772c7ef1 then? If we want to keep v1 I guess there is no need to remove that code?

I'm not against removing v1 though. But at the same time I'd like to increase major number of the plugin as this is fairly big change.

twaugh:
Just to clarify something I've sent mixed signals about before: there will still be 'docker save' tarballs with v2-only builds. But now, this plugin will no longer care whether there is or isn't a tarball, it only cares about OSBS logs.

The expectation for this pull request is that it will leave the entire procedure for creating a Koji Build up to OSBS.

When atomic-reactor finishes the build it will use the Content Generator API to create a Koji Build holding the metadata, the 'docker save' tarball and the logs.

That means this plugin should no longer:
* create a Koji Build
* fetch the 'docker save' tarball
* do anything with rpm-packages
* do anything with 'docker build' logs
* do anything special to handle scratch builds
* care about or have any knowledge about v1 vs v2 -- that comes from osbs.conf

maxamillion:
@twaugh - I'm pretty sure I got it all cleaned up, please let me know if there's anything else you'd like to see changed or removed. Thanks! :)

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -255,31 +197,6 @@ def __init__(self, id, method, params, session, options, workdir=None):
...
-                                                        follow=False)
-        self.logger.debug("Docker logs from OSBS retrieved")
-        outfile = open(build_log, 'w')
-        outfile.write(build_log_contents)
-        outfile.close()
-        file_names.append(fn)
-
-        return file_names
 
     def _get_file_url(self, source_filename):
```
twaugh: Pretty sure `_get_file_url` is no longer needed.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -301,35 +218,6 @@ def _getOutputImageTemplate(self, arch, compressed_extension=None):
 
         return "image-%s.tar%s" % (arch, compressed_extension)
```
twaugh: This looks like it's no longer needed.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -301,35 +218,6 @@ def _getOutputImageTemplate(self, arch, compressed_extension=None):
...
-                                                 ssl_verify_peer=ssl_verify_peer,
-                                                 ssl_verify_host=ssl_verify_host)
-        except urlgrabber.grabber.URLGrabError, error:
-            self.logger.info("Failed to download file from URL %s: %s.",
-                             remote_url, error)
-            return []
-        self.logger.debug("Output: %s.", output_filename)
-        return [output_filename]
 
     def nfs_dest_dir(self):
```
twaugh: Don't think we need this now.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -351,101 +239,10 @@ def osbs(self):
...
-
-    def getUploadPath(self):
-        """Get the path that should be used when uploading files to
-        the hub."""
-        return koji.pathinfo.taskrelpath(self.id)
 
     def resultdir(self):
         return os.path.join(self.workdir, 'osbslogs')
 
-    def _incremental_upload_logs(self, child_pid):
```
twaugh: However, we still need to make it possible to follow logs, and I think that's what this method is for.

maxamillion: I had previously removed the code that was using it and nobody seemed concerned so it was basically dead code and I pulled it out, but you're probably right and we'll just need to figure out how best to handle that now.

twaugh: Right, let me point out the bit that uses it which we need to keep...

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -499,87 +297,14 @@ def handler(self, src, target_info, arch, output_template, scratch=False,
...
-                    retry += 1
-                    time.sleep(10)
-                    continue
-                break
-            else:
-                self.logger.info("Gave up trying to save incremental logs "
-                                 "after #%d retries.", retry)
-                os._exit(1)
-            os._exit(0)
-
```
twaugh: The whole chunk above, which uses _incremental_upload_logs() and _write_incremental_logs(), needs to stay. This is what makes watching build logs work.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -499,87 +297,14 @@ def handler(self, src, target_info, arch, output_template, scratch=False,
...
         response = self.osbs().wait_for_build_to_finish(build_id)
         self.logger.debug("OSBS build finished with status: %s. Build "
                           "response: %s.", response.status,
                           response.json)
-        logs = self._download_logs(build_id)
```
twaugh: This is a separate fetch of the logs (for 'openshift-final.log'), which I don't think we ever really needed -- we already have the logs that were streamed from the build, and they're what we want.

twaugh:
Great!

There are still a couple more bits that are no longer needed:
* `CreateContainerTask._getOutputImageTemplate()`
* `BuildContainerTask._getOutputImageTemplate()`
* `BulidContainerTask._get_compressed_extension()`

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -482,6 +288,7 @@ def handler(self, src, target_info, arch, output_template, scratch=False,
             'target': target_info['name'],
             'architecture': arch,
             'yum_repourls': yum_repourls,
+            'scratch': scratch
```
twaugh: Tiny nit: could add a trailing comma at the end, just so any future additions are easy. Won't hold off merging for this though.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -546,40 +353,10 @@ def handler(self, src, target_info, arch, output_template, scratch=False,
...
-        rpmlist = []
-        if response.is_succeeded():
-            rpmlist = self._get_rpm_packages(response)
-
-        self.logger.debug("rpm list:")
-        for rpm in rpmlist:
-            self.logger.debug("%(name)s-%(version)s-%(release)s.%(arch)s.rpm" %
-                              rpm)
-
         repo_info = self.session.getRepo(target_info['build_tag'])
```
twaugh: Looks like this line is no longer needed now, sorry I missed this.

maxamillion: I actually think that was out previously and somehow snuck in because I missed it during a rebase. Will fix.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -591,30 +366,13 @@ def handler(self, src, target_info, arch, output_template, scratch=False,
...
-                              "there are %d: %s", len(files), files)
-        else:
-            for filename in files:
-                full_path = os.path.join(self.workdir, filename)
-                self.uploadFile(full_path, remoteName=output_template)
-                containerdata['files'].append(os.path.basename(output_template))
-
         return containerdata
 
-
```
twaugh: This is a stupid thing, but PEP8 says we should have two blank lines between classes. Won't hold up the merge for it though.

twaugh:
Looks good to me.

PR #37, opened 2016-04-27
Expose Koji CG build ID in CreateContainerTask

lcarva:


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -340,6 +340,12 @@ def _get_repositories(self, response):
                               repo_dict, type(error), error)
         return repositories
 
+    def _get_koji_build_id(self, response):
+        if hasattr(response, "get_koji_build_id"):
+            return response.get_koji_build_id()
+        else:
+            self.logger.info("Koji content generator build ID not available.")
```
vrutkovs: We should return something here, None or '' would work, probably

lcarva: In Python, None is returned when a `return` statement is not given:

```
>>> def foo():
...   pass
... 
>>> foo() is None
True
```

twaugh: `None` is probably best (i.e. make that explicit).

I'll note that `get_koji_build_id()` might return either `None` (that build didn't have an annotation for it) or '' (annotation was empty). It's fine, just something to be aware of.

vrutkovs: Right, although 'Explicit is better than implicit' :)

twaugh:
Looks good. I was wondering about whether we also want the string returned by the BuildContainerTask's handler() method to include a note of the koji Build IDs.

I'm still unclear on whether that string gets shown to the user who submitted the job at the command line.

lcarva:
Updated the cli portion of this plugin to display the URLs for any CGImport builds that were created. Note: This only happens if the command was executed in "wait" mode. Otherwise, results can't be displayed since tasks are still running.

koji_containerbuild/cli.py
```diff
@@ -33,6 +34,18 @@
...
 DEFAULT_CHANNEL = 'container'
 
+def print_task_result(task_id, result, weburl):
+    try:
+        result = json.loads(result)
+        result["koji_builds"] = ["%s/buildinfo?buildID=%s" % (weburl, build_id)
+                                 for build_id in result.get("koji_builds", [])]
+        result = json.dumps(result, sort_keys=True, indent=4)
+    except:
+        print "Unable to convert result output to JSON."
```
twaugh: I would use print() for compatibility with both Python 2 and Python 3.

lcarva: Remaining of file uses old-style print. Changing here requires changing for the whole file.

breillyrh:
Looks good to me.

PR #38, opened 2016-05-13
Don't start the build if package with this NVR already has been built

vrutkovs:
See also https://github.com/projectatomic/atomic-reactor/pull/485 - a similar check in atomic-reactor

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -237,8 +237,7 @@ class LabelsWrapper(object):
 
     def get_expected_nvr(self):
         data = self.get_data_labels()
-        name = data['NAME'].split('/')[-1]
-        return "{0}-{1}-{2}".format(name, data['VERSION'], data['RELEASE'])
+        return "{0}-{1}-{2}".format(data['COMPONENT'], data['VERSION'], data['RELEASE'])
 
     def format_label(self, label_id):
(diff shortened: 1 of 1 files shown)


twaugh:
Looking in more detail at this I think we might be using the wrong name in atomic reactor. We ought to be using the BZComponent label instead of the Name label.

twaugh:
Compare with master branch to see how this was did prior to CG. We should do the nvr check in the same place we used to initiate the Build object.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -619,6 +624,16 @@ def handler(self, src, target, opts=None):
                                       "target": target}
             if not SCM.is_scm_url(src):
                 raise koji.BuildError('Invalid source specification: %s' % src)
+            # Scratch builds shouldn't be checked for nvr
+            if not self.opts.get('scratch'):
+                expected_nvr = labels_wrapper.get_expected_nvr()
+                try:
+                    build_id = self.xmlrpc.getBuild(expected_nvr)['id']
+                except:
+                    self.logger.info("No build for %s found" % expected_nvr)
```
twaugh: Could change "%" to "," to use lazy string formatting provided by logging functions.

vrutkovs: Will do

twaugh:
LGTM

breillyrh:
Looks good to me as well, I'll merge.

PR #39, opened 2016-05-16
no need to warn about build result not being JSON

twaugh:
There is no need to warn if the build result cannot be parsed as JSON. This can happen if the CLI package is installed but the Koji builder has an older version of the koji-containerbuild-builder plugin.


twaugh:
CC @lcarva

lcarva:
LGTM

breillyrh:
Looks good to me as well, I'll merge.

PR #40, opened 2016-05-18
Supply koji_task_id to create_build()

twaugh:
This allows osbs-client to set a label on the BuildConfig which we can query later.

This parameter will be ignored with older osbs-client versions.


breillyrh:
LGTM, I'll merge.

PR #41, opened 2016-05-20
Handle release parameter

lcarva:
Needs https://github.com/projectatomic/osbs-client/pull/424.

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -55,7 +55,6 @@ finally:
     fo.close()
 
 
-
 # List of LABEL identifiers used within Koji. Values doesn't need to correspond
 # to actual LABEL names. All these are required unless there exist default
 # value (see LABEL_DEFAULT_VALUES).
@@ -176,11 +175,12 @@ class FileWatcher(object):
 
 
 class LabelsWrapper(object):
-    def __init__(self, dockerfile_path, logger_name=None):
+    def __init__(self, dockerfile_path, logger_name=None, labels_override=None):
         self.dockerfile_path = dockerfile_path
         self._setup_logger(logger_name)
         self._parser = None
         self._label_data = {}
+        self._labels_override = labels_override or {}
 
     def _setup_logger(self, logger_name=None):
         if logger_name:
@@ -193,7 +193,10 @@ class LabelsWrapper(object):
     def get_labels(self):
         """returns all labels how they are found in Dockerfile"""
         self._parse()
-        return self._parser.labels
+        labels = self._parser.labels
+        if self._labels_override:
+            labels.update(self._labels_override)
+        return labels
 
     def get_data_labels(self):
         """Subset of labels found in Dockerfile which we are interested in
@@ -258,7 +261,6 @@ class CreateContainerTask(BaseTaskHandler):
         BaseTaskHandler.__init__(self, id, method, params, session, options,
                                  workdir)
         self._osbs = None
-
 
     def osbs(self):
         """Handler of OSBS object"""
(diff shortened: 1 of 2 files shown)


twaugh:
`LabelsWrapper.get_labels()` needs to use the supplied labels in preference to those in the Dockerfile.

lcarva:
@twaugh The "release" value is only read as a task opt parameter, not from the Dockerfile. From a koji-containerbuild POV, it doesn't know that this will ultimately be turned into a LABEL by osbs-client.

twaugh:
Are you sure? I think e.g. `LabelsWrapper().get_data_labels()` will complain if it doesn't see a Release label in the Dockerfile, and `LabelsWrapper().get_expected_nvr()` will use Release from the Dockerfile (only) to construct the n-v-r, etc.

lcarva:
@twaugh You're right. Added the concept of `labels_override` to `LabelsWrapper` to address this concern.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -639,7 +655,9 @@ def handler(self, src, target, opts=None):
                                      opts.get('scratch', False),
                                      opts.get('yum_repourls', None),
                                      opts.get('git_branch', None),
-                                     opts.get('push_url', None))
+                                     opts.get('push_url', None),
+                                     labels_override,
```
twaugh: This could be `labels=labels_override` to make it safer against future changes to arguments.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -606,7 +620,9 @@ def handler(self, src, target, opts=None):
         archlist = self.getArchList(build_tag)
 
         dockerfile_path = self.fetchDockerfile(src)
-        labels_wrapper = LabelsWrapper(dockerfile_path, self.logger.name)
+        labels_override = self._get_labels_override(opts)
+        labels_wrapper = LabelsWrapper(dockerfile_path, self.logger.name,
+                                       labels_override)
```
twaugh: This could be `labels_override=labels_override`.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -520,7 +528,8 @@ def runBuilds(self, src, target_info, arches, scratch=False,
                                                                 scratch,
                                                                 yum_repourls,
                                                                 branch,
-                                                                push_url],
+                                                                push_url,
+                                                                labels],
```
twaugh: Is the default `None` value valid here? If not, we should use something like `labels or {}`.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -176,11 +175,12 @@ def clean(self):
 
 
 class LabelsWrapper(object):
-    def __init__(self, dockerfile_path, logger_name=None):
+    def __init__(self, dockerfile_path, logger_name=None, labels_override=None):
         self.dockerfile_path = dockerfile_path
         self._setup_logger(logger_name)
         self._parser = None
         self._label_data = {}
+        self._labels_override = labels_override
```
twaugh: This should be `labels_override or {}` I think.

koji_containerbuild/plugins/builder_containerbuild.py
lcarva: Yes. `CreateContainerTask.handler` takes care of `None`.

koji_containerbuild/plugins/builder_containerbuild.py
lcarva: It's not needed, since `self._labels_override` is only used if truthy (either None or {} will be falsy). But defaulting to `{}` seems safer.

twaugh: Actually I think default to None is better than default to {}. Better to have immutable types for method argument defaults.

twaugh: "truthy", haven't heard that before :)

koji_containerbuild/plugins/builder_containerbuild.py
lcarva: Done.

koji_containerbuild/plugins/builder_containerbuild.py
lcarva: Done.

twaugh:
LGTM

koji_containerbuild/plugins/builder_containerbuild.py
lcarva: My javascript background spills over every now and then :)

Method argument defaults are shared between calls. In most cases, this is not the intention and can cause some nasty bugs - thus immutable values are better suited. My approach here was to default to a value within the function, which creates a new object for every function call.

PR #42, opened 2016-05-27
make streamed logs line-buffered

twaugh:
Currently, developers cannot follow build logs line-by-line because the incremental build log is only written to disk when the write buffer is full.

This commit flushes the output after each line.

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -322,6 +322,7 @@ class CreateContainerTask(BaseTaskHandler):
         try:
             for line in build_logs:
                 outfile.write("%s\n" % line)
+                outfile.flush()
         except Exception, error:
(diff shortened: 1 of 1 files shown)


vrutkovs:
LGTM

pbabinca:
Same code is present in upstream Koji. Which could mean developers decided it is too disruptive to transfer so small chunks of data around. Or developers who use other builds in Koji may benefit of the same fix.

twaugh:
I'm pretty sure it works fine for RPM builds, just not for container builds.

twaugh:
So for builder/kojid, it executes mock in the equivalent portion to this code. mock writes its logs in a line-buffered manner because it uses the logging module.

I think we need to flush() after each write in order to have that same behaviour.

vrutkovs:
Verified on local koji - log lines are no longer broken in the middle and appear faster

lcarva:
LGTM

PR #43, opened 2016-06-01
Fix task result output

lcarva:


koji_containerbuild/cli.py
```diff
@@ -35,18 +35,32 @@
 DEFAULT_CHANNEL = 'container'
 
 
+def print_value(value, level, indent, suffix=''):
+    offset = ' ' * level * indent
+    print ''.join([offset, str(value), suffix])
+
+
+def print_result(result, level=0, indent=2):
+    if type(result) == list:
```
twaugh: Should be `if isinstance(result, list):`

koji_containerbuild/cli.py
```diff
@@ -35,18 +35,32 @@
...
+        for item in result:
+            print_result(item, level+1)
+    elif type(result) == dict:
```
twaugh: Should be 'if isinstance(result, dict):`

koji_containerbuild/cli.py
```diff
@@ -35,18 +35,32 @@
...
                                  for build_id in result.get("koji_builds", [])]
-        result = json.dumps(result, sort_keys=True, indent=4)
-    except ValueError:
-        # Unable to convert result output to JSON
+    except TypeError:
         pass
 
     print "Task Result (%s):" % task_id
-    print result
+    print_result(result)
```
twaugh: Before the builder is upgraded, result might be a string here, correct? I think we should not print anything in that case.

lcarva: It will print it as a string by the `else` statement in print_result function.

koji_containerbuild/cli.py
lcarva: yes. it's safer to do so, but I doubt koji will extend `list` and `dict` classes.

koji_containerbuild/cli.py
twaugh: ok

twaugh:
LGTM

breillyrh:
LGTM as well, merging.

PR #45, opened 2016-06-08
Pass correct param to osbs-client with --release option

vrutkovs:
Fixes several typos after #41 was merged


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -389,7 +389,7 @@ def handler(self, src, target_info, arch, scratch=False,
         if push_url:
             create_build_args['git_push_url'] = push_url
         if labels:
-            create_build_args['labels'] = labels
+            create_build_args['label'] = labels
```
twaugh: No, 'labels' is correct.

vrutkovs:
My fault, it seems the problem is in osbs-client

PR #46, opened 2016-06-08
Should use label 'Release' to set release

twaugh:


vrutkovs:
LGTM, tested on local koji instance

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -605,7 +605,7 @@ def _get_labels_override(self, opts):
         labels = {}
         release = opts.get('release', None)
         if release:
-            labels['release'] = release
+            labels['Release'] = release
```
lcarva: Remove lowercase `'release'` from `LABEL_NAME_MAP`? https://github.com/twaugh/koji-containerbuild/blob/6189ec047686f45f8dd43c807e5b07e705158af0/koji_containerbuild/plugins/builder_containerbuild.py#L75

twaugh: No, I think the rule until we switch to the new label set completely is:
* prefer the new set when reading labels
* use the old set when writing labels

lcarva:
LGTM

PR #50, opened 2016-06-21
Fail createContainer if OpenShift build fails

### LINKED ISSUE #49: Check build result in createContainer
Currently createContainer is marked as _closed_, even when the OpenShift build fails. Then buildContainer checks if the build was successful, which is its only interaction with OpenShift.

Can the check be moved to createContainer so that it's marked as _failed_ when the build fails? The current behaviour is a bit confusing.

Furthermore, if only one of those tasks interacted with OpenShift it would unbreak the use case where koji builders in single channel talk to different OpenShift instances.

Related to #8.

mmilata:
Closes #49.

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -453,6 +453,10 @@ class CreateContainerTask(BaseTaskHandler):
 
         self.logger.info("Response status: %r", response.is_succeeded())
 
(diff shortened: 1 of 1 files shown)


vrutkovs:
Tested on local koji instance - works as expected

lcarva:
LGTM

breillyrh:
LGTM, I'll merge.

PR #52, opened 2016-06-23
Disable scratch builds

### LINKED ISSUE #51: Disable scratch builds until a proper solution is found
Currently scratch builds are overwriting the normal builds, so we should disallow scratch builds and output a warning for now

vrutkovs:
Fixes #51

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -619,6 +619,9 @@ class BuildContainerTask(BaseTaskHandler):
         self.opts = opts
         data = {}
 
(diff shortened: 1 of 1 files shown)


twaugh:
Didn't we discover that cli.py isn't really used?

I think we want to fail the task from within the builder plugin.

lcarva:
LGTM

That's correct, `cli.py` is only used when running `koji container-build` from the command line. When the task is initiated via koji's API, it bypasses `cli.py`. I see the check is now done in `BuildContainerTask.handle()` method which seems correct.

breillyrh:
LGTM, merging.

PR #55, opened 2016-06-23
Show build's error message if available

vrutkovs:
See https://github.com/projectatomic/osbs-client/pull/444

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -356,6 +356,15 @@ class CreateContainerTask(BaseTaskHandler):
 
         return koji_build_id
 
+    def _get_error_message(self, response):
+        error_message = None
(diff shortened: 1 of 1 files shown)


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -476,7 +490,7 @@ def handler(self, src, target_info, arch, scratch=False,
             'osbs_build_id': build_id,
             'files': [],
             'repositories': repositories,
-            'koji_build_id': koji_build_id,
+            'koji_build_id': koji_build_id
```
twaugh: Generally it's nice to leave a terminating "," in a list or dict, so that adding another line at the end doesn't trip up by forgetting to do it.

vrutkovs: Right, added a comma

lcarva:
LGTM

PR #57, opened 2016-07-05
Drop support for release parameter

lcarva:
Verified that `buildContainer` does not complain about accepting "release" parameter, but it's completely ignored and not forwarded to `createContainer` task.


twaugh:
LGTM

PR #60, opened 2016-10-18
Cancel build when createContainer task is cancelled (throws koji.GenericException)

vrutkovs:
Trap `koji.ActionNotAllowed` exception - it is thrown when we try to incrementally write a log and koji task is dead - which means its cancelled.

This works when `createContainer` task is cancelled, not `buildContainer`. Tested on local koji.

PTAL @twaugh @lcarva @breillyrh

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -413,7 +413,10 @@ class CreateContainerTask(BaseTaskHandler):
         koji.ensuredir(osbs_logs_dir)
         pid = os.fork()
         if pid:
-            self._incremental_upload_logs(pid)
+            try:
+                self._incremental_upload_logs(pid)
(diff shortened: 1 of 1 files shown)


twaugh:
What happens when the `buildContainer` task is cancelled?

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -413,7 +413,10 @@ def handler(self, src, target_info, arch, scratch=False,
         koji.ensuredir(osbs_logs_dir)
```
vrutkovs: Openshift cancels it correctly, tested it

vrutkovs:
> What happens when the buildContainer task is cancelled?

Koji simply kills the plugin's python instance and no exceptions thrown - so this approach won't work. This should be resolved by merging the tasks

twaugh:
LGTM, nice work!

PR #62, opened 2016-10-19
tests: make tests work

twaugh:
Add tests/\__init__.py so we test the in-tree module.
Mock kojid.


twaugh:
@vrutkovs

lcarva:
we should consider using MagicMock in the future: https://github.com/testing-cabal/mock

PR #63, opened 2016-10-19
Cancel OSE builds when koji task is cancelled Mk.2

vrutkovs:
Use SIGINT signal handler to cancel OSE build (if it has been started). This change 
ensures that the build in New/Pending state (where logs are not being written) is 
also correctly cancelled.

Tested on local koji instance

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -33,6 +33,7 @@ import urlgrabber
 import urlgrabber.grabber
 import dockerfile_parse
 import pycurl
+import signal
 
 import koji
 from koji.daemon import SCM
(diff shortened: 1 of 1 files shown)


vrutkovs:
PTAL @twaugh @lcarva @pbabinca

lcarva:
Verified that this works as expected.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -406,11 +409,16 @@ def handler(self, src, target_info, arch, scratch=False,
         build_id = build_response.get_build_name()
         self.logger.debug("OSBS build id: %r", build_id)
 
-        self.logger.debug("Waiting for osbs build_id: %s to be scheduled.",
-                          build_id)
-        # we need to wait for kubelet to schedule the build, otherwise it's 500
-        self.osbs().wait_for_build_to_get_scheduled(build_id)
```
twaugh: Are we no longer waiting for the build to get scheduled?

vrutkovs: Its no longer necessary - OSE doesn't throw HTTP 500 when logs are fetched for Pending build

twaugh: I wonder if Origin behaves the same way. Don't want to tie this to OSE specifically.

vrutkovs: Didn't try this on Origin, so not sure which version is required for that to work. I'll revert that part, and make sure that `signal.signal` solution works anyway

twaugh:
LGTM from code inspection but I haven't tested it.

PR #64, opened 2016-10-20
Reject a build if the longest tag is longer than 128 chars

vrutkovs:


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -691,6 +691,15 @@ def handler(self, src, target, opts=None):
                     raise koji.BuildError(
                         "Build for %s already exists, id %s" % (expected_nvr, build_id))
 
+            # Make sure the longest tag for the docker image is less than 128 chars
```
twaugh: 'no more than'

vrutkovs: Fixed, 128 chars is allowed

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -230,6 +230,16 @@ def get_extra_data(self):
             extra_data[extra_key] = value
         return extra_data
 
+    def get_additional_tags(self):
+        """Returns a list of additional tags to be applied to an image"""
+        tags = []
+        additional_tags_path = os.join(self.dockerfile_path, '..', 'additional-tags')
```
lcarva: You probably want to extract the dirname from self.dockerfile_path, no?

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -691,6 +702,17 @@ def handler(self, src, target, opts=None):
                     raise koji.BuildError(
                         "Build for %s already exists, id %s" % (expected_nvr, build_id))
 
+            # Make sure the longest tag for the docker image is no more than 128 chars
+            # see https://github.com/docker/docker/issues/8445
+
+            version_release_tag = "%s-%s" % (LABEL_DATA_MAP['VERSION'], LABEL_DATA_MAP['RELEASE'])
+            tags = labels_wrapper.get_additional_tags() + version_release_tag
```
lcarva: Isn't this an invalid operation? `list` + `str`
Why not use append?

vrutkovs: Right, fixed

koji_containerbuild/plugins/builder_containerbuild.py
vrutkovs: Agreed, its a cleaner solution

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -230,6 +230,17 @@ def get_extra_data(self):
...
+    def get_additional_tags(self):
+        """Returns a list of additional tags to be applied to an image"""
+        tags = []
+        dockerfile_dir = os.path.dirname(self.dockerfile_path)
+        additional_tags_path = os.join(dockerfile_dir, 'additional-tags')
+        try:
+            with open(additional_tags_path, 'r') as fd:
+                tags.append(fd.readline())
+        finally:
+            return tags
```
twaugh: We ought to strip these lines. Something like this?

```
try:
    with open(...) as fd:
        return [tag.strip() for tag in fd]
except IOError:
    return []
```

vrutkovs: Fixed. Tags with '-' are skipped (similar to atomic-reactor's behaviour)

PR #66, opened 2017-03-01
Use osbs-client's create_orchestrator_build method if available

vrutkovs:
Related to https://github.com/projectatomic/osbs-client/pull/527

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -388,21 +388,38 @@ class BuildContainerTask(BaseTaskHandler):
         elif pkg_cfg['blocked']:
             raise koji.BuildError("package (container)  %s is blocked for tag %s" % (name, target_info['dest_tag_name']))
 
+    def has_orchestrator(self):
+        method = getattr(self.osbs(), "create_orchestrator_build", None)
+        return callable(method)
+
     def runBuilds(self, src, target_info, arches, scratch=False,
                   yum_repourls=None, branch=None, push_url=None):
 
         self.logger.debug("Spawning jobs for arches: %r" % (arches))
 
         results = []
-        for arch in arches:
-            result = self.createContainer(
-                src, target_info, arch, scratch, yum_repourls, branch, push_url)
-            results.append(result)
+
+        kwargs = dict(
+            src=src,
+            target_info=target_info,
+            scratch=scratch,
+            yum_repourls=yum_repourls,
+            branch=branch,
+            push_url=push_url
+        )
+
+        if self.has_orchestrator():
+            kwargs["arches"] = arches
+        else:
+            kwargs["arch"] = arches[0]
+
+        results = [self.createContainer(**kwargs)]
 
         self.logger.debug("Results: %r", results)
         return results
 
-    def createContainer(self, src, target_info, arch, scratch, yum_repourls, branch, push_url):
+    def createContainer(self, src=None, target_info=None, arch=None, arches=None,
+                        scratch=None, yum_repourls=[], branch=None, push_url=None):
         if not yum_repourls:
             yum_repourls = []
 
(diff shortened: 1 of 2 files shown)


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -388,21 +388,39 @@ def check_whitelist(self, name, target_info):
...
+            yum_repourls=yum_repourls,
+            branch=branch,
+            push_url=push_url
+        )
+
+        if self.has_orchestrator():
+            kwargs["arches"] = arches
+            results = [self.createContainer(**kwargs)]
+        else:
+            for arch in arches:
```
lcarva: Even though this is the existing behavior, it's not correct. This could be even more problematic as we transition over to fully support multiple architectures.

What if we simply log an error and pick the first item in `arches`? I think just ignoring the value of `arches` altogether and setting it to `'x86_64'` is also fine.

lcarva: Or just not set `arch` at all in this case.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -427,13 +445,21 @@ def createContainer(self, src, target_info, arch, scratch, yum_repourls, branch,
             'scratch': scratch,
             'koji_task_id': self.id,
         }
+        if arch:
+            create_build_args['platform'] = arch
+            create_method = self.osbs().create_build
+        elif arches:
+            create_build_args['platformes'] = arch
```
lcarva: Typo: `platforms`

lcarva:
@vrutkovs @twaugh should we have a config option in order to enable usage of `create_orchestrator_build`? Not sure if Fedora cares.

koji_containerbuild/plugins/builder_containerbuild.py
vrutkovs: Fixed, thanks

koji_containerbuild/plugins/builder_containerbuild.py
vrutkovs: Agree, I'll replace this with first available item - and leave the PR in WIP until this is tested properly

twaugh:
Actually my idea was that we would cut a new major release of koji-containerbuild for this change. Rollback would be to downgrade to previous version.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -388,21 +388,38 @@ def check_whitelist(self, name, target_info):
...
+            scratch=scratch,
+            yum_repourls=yum_repourls,
+            branch=branch,
+            push_url=push_url
+        )
+
+        if self.has_orchestrator():
+            kwargs["arches"] = arches
+        else:
+            kwargs["arch"] = arches[1]
```
twaugh: not `[0]`?

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -427,13 +444,21 @@ def createContainer(self, src, target_info, arch, scratch, yum_repourls, branch,
             'scratch': scratch,
             'koji_task_id': self.id,
         }
+        if arch:
+            create_build_args['platform'] = arch
+            create_method = self.osbs().create_build
+        elif arches:
```
twaugh: Could put the 'arches' test before the 'arch' test to make it clear `create_orchestrator_build` is the preferred method and the old method is there only for compatibility.

koji_containerbuild/plugins/builder_containerbuild.py
vrutkovs: Right, its a typo, fixed

koji_containerbuild/plugins/builder_containerbuild.py
vrutkovs: Sure, fixed

vrutkovs:
Tested on internal koji and osbs-dev. K-cb parts have worked flawlessly, but some more modifications might be required

lcarva:
@vrutkovs please cherry-pick this unit testing commit: https://github.com/release-engineering/koji-containerbuild/commit/c8296ac0aa85797dc8ee52d9f6c66090d679fc46

vrutkovs:
@lcarva pushed, thanks!

PR #67, opened 2017-03-23
orchestrator build can only successfully run if can_orchestrate

rcerven:
option is set, in osbs.conf
otherwise will fallback to regular prod_build

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -39,6 +39,7 @@ from koji.tasks import ServerExit, BaseTaskHandler
 
 from osbs.api import OSBS
 from osbs.conf import Configuration
+from osbs.exceptions import OsbsValidationException
 
 # We need kojid module which isn't proper python module and not even in
 # site-package path.
@@ -388,10 +389,6 @@ class BuildContainerTask(BaseTaskHandler):
         elif pkg_cfg['blocked']:
             raise koji.BuildError("package (container)  %s is blocked for tag %s" % (name, target_info['dest_tag_name']))
 
-    def has_orchestrator(self):
-        method = getattr(self.osbs(), "create_orchestrator_build", None)
-        return callable(method)
-
     def runBuilds(self, src, target_info, arches, scratch=False,
                   yum_repourls=None, branch=None, push_url=None):
 
@@ -405,20 +402,16 @@ class BuildContainerTask(BaseTaskHandler):
             scratch=scratch,
             yum_repourls=yum_repourls,
             branch=branch,
-            push_url=push_url
+            push_url=push_url,
+            arches=arches
         )
-
-        if self.has_orchestrator():
-            kwargs["arches"] = arches
-        else:
-            kwargs["arch"] = arches[0]
 
         results = [self.createContainer(**kwargs)]
 
         self.logger.debug("Results: %r", results)
         return results
 
-    def createContainer(self, src=None, target_info=None, arch=None, arches=None,
+    def createContainer(self, src=None, target_info=None, arches=None,
                         scratch=None, yum_repourls=[], branch=None, push_url=None):
         if not yum_repourls:
             yum_repourls = []
@@ -432,6 +425,10 @@ class BuildContainerTask(BaseTaskHandler):
         scm.assert_allowed(self.options.allowed_scms)
         git_uri = scm.get_git_uri()
         component = scm.get_component()
+        arch = None
+
+        if not arches:
+            raise ContainerError("arches aren't specified")
 
         create_build_args = {
             'git_uri': git_uri,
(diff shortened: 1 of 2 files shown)


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -444,22 +436,26 @@ def createContainer(self, src=None, target_info=None, arch=None, arches=None,
...
         if push_url:
             create_build_args['git_push_url'] = push_url
 
         self.logger.debug("Starting %s with params: '%s'", create_method, create_build_args)
-        build_response = create_method(**create_build_args)
+
+        try:
+            self.logger.debug("Starting create_orchestrator_build with params: '%s, platforms=%r'",
+                              create_build_args, arches)
+            build_response = self.osbs().create_orchestrator_build(**create_build_args, platforms=arches)
```
twaugh: I think you need to create a copy of this dict and update it.

There's a py.test suite, it just doesn't get run automatically.

vrutkovs: `create_build_args` and `create_build_method` is probably a cleaner way to do this

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -444,22 +436,31 @@ def createContainer(self, src=None, target_info=None, arch=None, arches=None,
...
             create_build_args['git_push_url'] = push_url
 
         self.logger.debug("Starting %s with params: '%s'", create_method, create_build_args)
-        build_response = create_method(**create_build_args)
+
+        try:
+            create_build_args['platforms'] = arches
+            create_method = self.osbs().create_orchestrator_build
+            self.logger.debug("Starting %s with params: '%s'",
+                              create_method, create_build_args)
```
twaugh: Does this get logged twice? See L447.

rcerven: yeah removed from 447

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -432,34 +425,43 @@ def createContainer(self, src=None, target_info=None, arch=None, arches=None,
...
         if branch:
             create_build_args['git_branch'] = branch
         if push_url:
             create_build_args['git_push_url'] = push_url
 
-        self.logger.debug("Starting %s with params: '%s'", create_method, create_build_args)
-        build_response = create_method(**create_build_args)
+        try:
+            create_build_args['platforms'] = arches
+            create_method = self.osbs().create_orchestrator_build
```
lcarva: Don't think we need to store a reference to the method anymore. Can't we just call it instead?

Same comment for `self.osbs().create_build` below.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -432,34 +425,43 @@ def createContainer(self, src=None, target_info=None, arch=None, arches=None,
...
+            self.logger.debug("Starting %s with params: '%s'",
+                              create_method, create_build_args)
+            build_response = create_method(**create_build_args)
+        except (AttributeError, OsbsValidationException):
+            # Older osbs-client, or else orchestration not enabled
+            create_build_args['platform'] = arches[0]
+            create_build_args['architecture'] = arch = arches[0]
```
lcarva: Nitpick: IMO, these two lines would be easier to read with something like:

```python
arch = arches[0]
create_build_args['platform'] = arch
create_build_args['architecture'] = arch
```

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -432,34 +425,43 @@ def createContainer(self, src=None, target_info=None, arch=None, arches=None,
...
-
         if branch:
             create_build_args['git_branch'] = branch
         if push_url:
             create_build_args['git_push_url'] = push_url
 
-        self.logger.debug("Starting %s with params: '%s'", create_method, create_build_args)
-        build_response = create_method(**create_build_args)
+        try:
+            create_build_args['platforms'] = arches
```
lcarva: Let's do this instead of updating the `create_build_args` dict:

```python
build_response = self.osbs().create_orchestrator_build(platforms=arches, **create_build_args)
```

This will avoid accidentally passing `platforms` to `create_build` method.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -432,34 +425,43 @@ def createContainer(self, src=None, target_info=None, arch=None, arches=None,
...
+            create_method = self.osbs().create_orchestrator_build
+            self.logger.debug("Starting %s with params: '%s'",
+                              create_method, create_build_args)
+            build_response = create_method(**create_build_args)
+        except (AttributeError, OsbsValidationException):
+            # Older osbs-client, or else orchestration not enabled
+            create_build_args['platform'] = arches[0]
```
twaugh: I actually don't think it's correct to set `platform` for the regular build. This isn't a worker build.

We've only been setting this since commit 63ece7e6708b006fbdefdbe47ad21ce68680fb01, and I think that mistake wasn't caught in review at the time.

koji_containerbuild/plugins/builder_containerbuild.py
twaugh: (But see https://github.com/release-engineering/koji-containerbuild/pull/67/files#r107860016 before addressing this)

koji_containerbuild/plugins/builder_containerbuild.py
rcerven: well yeah I had that before but that does work only in python3

koji_containerbuild/plugins/builder_containerbuild.py
rcerven: ahh ok wasn't aware of that, will remove that

koji_containerbuild/plugins/builder_containerbuild.py
twaugh: No, the '**kwargs' part is afterwards in Lui's example. This works in python2.6.

koji_containerbuild/plugins/builder_containerbuild.py
rcerven: well there is method call and as well used in logger, so rather than writing it twice, reference seems good

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -432,34 +425,40 @@ def createContainer(self, src=None, target_info=None, arch=None, arches=None,
...
         if push_url:
             create_build_args['git_push_url'] = push_url
 
-        self.logger.debug("Starting %s with params: '%s'", create_method, create_build_args)
-        build_response = create_method(**create_build_args)
+        try:
+            create_method = self.osbs().create_orchestrator_build
+            self.logger.debug("Starting %s with params: '%s, platforms:%s'",
+                              create_method, create_build_args, arches)
+            build_response = create_method(platforms=arches, **create_build_args)
```
vrutkovs: Agree with Tim and Lui on this - there is no point to use `create_method` if it would be used just once.

This might be a good idea to print its name in some error message though

PR #70, opened 2017-06-28
Cancel a koji task if the osbs task is cancelled

maxamillion:
Signed-off-by: Adam Miller <maxamillion@fedoraproject.org>

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -94,6 +94,10 @@ class ContainerError(koji.GenericError):
     """Raised when container creation fails"""
     faultCode = 2001
 
+class ContainerCancelled(koji.GenericError):
+    """Raised when container creation is cancelled by OSBS"""
+    faultCode = 2002
+
 
 # TODO: push this to upstream koji
 class My_SCM(SCM):
@@ -532,7 +536,12 @@ class BuildContainerTask(BaseTaskHandler):
 
         self.logger.info("Response status: %r", response.is_succeeded())
 
-        if response.is_failed():
+        if response.is_cancelled():
+            this_task.cancel()
+            # FIXME - What kind of information do we want in here?
+            raise ContainerCancelled('Image build was cancelled by OSBS')
+
+        elif response.is_failed():
             error_message = self._get_error_message(response)
             if error_message:
                 raise ContainerError('Image build failed. %s. OSBS build id: %s' %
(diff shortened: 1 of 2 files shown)


maxamillion:
Depends on https://github.com/projectatomic/osbs-client/pull/616

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -532,7 +536,12 @@ def sigint_handler(*args, **kwargs):
 
         self.logger.info("Response status: %r", response.is_succeeded())
 
-        if response.is_failed():
+        if response.is_cancelled():
+            this_task.cancel()
+            # FIXME - What kind of information do we want in here?
```
twaugh: I think this is as much as you can provide already, since we don't get to supply a reason when cancelling.

The only suggestion I have otherwise is to include a speculation in the message: " (maybe by automated rebuild?)".

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -94,6 +94,10 @@ class ContainerError(koji.GenericError):
     """Raised when container creation fails"""
     faultCode = 2001
 
+class ContainerCancelled(koji.GenericError):
+    """Raised when container creation is cancelled by OSBS"""
+    faultCode = 2002
```
lcarva: Out of curiosity, how do we know which faultCode to use?

maxamillion: I honestly don't know, I was digging through the koji code and the `faultCode` of `2001` doesn't appear to actually mean anything to koji so I just incremented since it appears to be arbitrary.

koji_containerbuild/plugins/builder_containerbuild.py
maxamillion: Yeah, I'll add the speculation message for good measure. +1

maxamillion:
On the topic of unit testing, I couldn't find any tests for any of the `BuildResponse` "status checkers" (`is_failed`, `is_running`, `is_pending`, etc). Should I just write new tests and go ahead and cover all of those for good measure?

twaugh:
Adding additional testing for the other state checkers would be really great if you'd like to. :)

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -532,7 +536,13 @@ def sigint_handler(*args, **kwargs):
 
         self.logger.info("Response status: %r", response.is_succeeded())
 
-        if response.is_failed():
+        if response.is_cancelled():
+            this_task.cancel()
```
twaugh: Is this right? Isn't `this_task` a dict?

twaugh:
Tests fail (run them with `py.test`).

koji_containerbuild/plugins/builder_containerbuild.py
maxamillion: Yes, this was an oversight on my part. For some reason I had it in my head that `this_task` was a koji `Task` instance. It is not. I'll get this fixed up.

PR #71, opened 2017-07-17
Update koji-containerbuild to support koji 1.13

vrutkovs:
* `activate_session` new requires two arguments
* `_running_in_bg` has moved to koji_cli.lib

Tested using fedora-based osbs-box

### DIFF
--- a/koji_containerbuild/cli.py
+++ b/koji_containerbuild/cli.py
@@ -30,7 +30,6 @@ from optparse import OptionParser
 # its functions. And this CLI don't necessary be named koji.
 clikoji = None
 
-
 # matches hub's buildContainer parameter channel
 DEFAULT_CHANNEL = 'container'
 
@@ -103,7 +102,23 @@ def handle_container_build(options, session, args):
         parser.error(_("Exactly two arguments (a build target and a SCM URL "
                        "or archive file) are required"))
         assert False
-    clikoji.activate_session(session)
+
+    # Koji API has changed - activate_session requires two arguments
+    # _running_in_bg has been moved to koji_cli.lib
+    try:
+        from koji_cli.lib import _running_in_bg, activate_session
+    except ImportError:
+        # Create wrappers for backwards compatibility.
+        _running_in_bg = clikoij._running_in_bg
+
+        def activate_session(session, options):
+            try:
+                clikoji.activate_session(session)
+            except TypeError:
+                clikoji.activate_session(session, options)
+    
+    activate_session(session, options)
+
     target = args[0]
     if target.lower() == "none" and build_opts.repo_id:
         target = None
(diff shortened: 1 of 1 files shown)


koji_containerbuild/cli.py
```diff
@@ -103,7 +103,16 @@ def handle_container_build(options, session, args):
         parser.error(_("Exactly two arguments (a build target and a SCM URL "
                        "or archive file) are required"))
         assert False
-    clikoji.activate_session(session)
+    # Koji API has changed - activate_session requires two arguments
+    try:
+        clikoji.activate_session(session)
+    except TypeError:
+        clikoji.activate_session(session, options)
```
lcarva: Looks like `activate_session` can now also be imported from `koji_cli.lib`. Let's use that here, too.
We can also perform this check earlier on, something like:

```python

clikoji = None

try:
  from koji_cli.lib import _running_in_bg, activate_session
except ImportError:
  # Create wrappers for backwards compatibility.
  _running_in_bg = clikoij._runing_in_bg

  def activate_session(session, options):
    try:
      clikoji.activate_session(session)
    except TypeError:
      clikoji.activate_session(session, options)
  
```

vrutkovs: Oh, sounds like a good idea, implemented that

koji_containerbuild/cli.py
```diff
@@ -103,7 +103,27 @@ def handle_container_build(options, session, args):
         parser.error(_("Exactly two arguments (a build target and a SCM URL "
                        "or archive file) are required"))
         assert False
-    clikoji.activate_session(session)
+
+    # Koji API has changed - activate_session requires two arguments
+    # _running_in_bg has been moved to koji_cli.lib
+    try:
```
lcarva: Maybe move this to the top of the file right after `clikoji = None` ?

lcarva: ping @vrutkovs

vrutkovs: Right, testing this

vrutkovs: Yes, this worked, updated the PR

vrutkovs: Had to roll this back - `clikoji` is not yet set by that time, so it doesn't have `_running_in_bg` option

lcarva: aha! Thanks for looking into it.

koji_containerbuild/cli.py
```diff
@@ -103,7 +102,23 @@ def handle_container_build(options, session, args):
...
+        from koji_cli.lib import _running_in_bg, activate_session
+    except ImportError:
+        # Create wrappers for backwards compatibility.
+        _running_in_bg = clikoij._running_in_bg
```
twaugh: koij?

vrutkovs: Created https://github.com/release-engineering/koji-containerbuild/pull/72 to fix that

PR #73, opened 2017-08-24
Add support for isolated builds

lcarva:


lcarva:
Remaining work:

- [ ] Integration testing
- [ ] osbs-client API support: https://github.com/projectatomic/osbs-client/pull/674

lcarva:
@twaugh, @vrutkovs, @rcerven, @bfontecc007: I just realized that koji-containerbuild will blindly attempt to fallback to `create_build` if `create_orchestrator_build` raises `OsbsValidationException`. This is really not ideal because various, unrelated, scenarios will raise this exception, i.e. kojihub missing, no username for pulp, etc.

This was made more obvious with the introduction of isolated builds, as we check if a release value was provided and it matches the expected format, raising `OsbsValidationException` otherwise. The usage of the exception in this case seem reasonable to me.

I'm not sure what the correct solution is here. A new exception, OsbsOrchestratorNotEnabled, in osbs-client makes sense. It might just be a bit tricky to keep backwards compatibility. (Try to import OsbsOrchestratorNotEnabled, if `ImportError` check `OsbsValidationException` intead?) Can also make the new exception inherit from the more generic one.

Thoughts?

bfontecc007:
What's wrong with just changing it back to RuntimeError?

lcarva:
> What's wrong with just changing it back to RuntimeError?

Because this is really only masking the issue.

bfontecc007:
Yeah if you want to clear up the other cases as well. Inheriting from the
more generic one sounds like it makes sense to me.

lcarva:
Added `OsbsOrchestratorNotEnabled` exception as its own commit. PTAL

PR #74, opened 2017-08-25
Flatpak support

owtaylor:
This builds upon https://github.com/projectatomic/atomic-reactor/pull/747 and https://github.com/projectatomic/osbs-client/pull/656. I chose to create a new `koji-containerbuild flatpak-build` rather than adding a `--flatpak` option to reduce the amount that the user had to worry about inter-option dependencies. The sharing of code between the two commands got a bit messy because there were some options for `container-build` that seem like leftovers (for compatibility?) and I didn't want to inherit them into `flatpak-build` - if that cleanup was done globally, there would be less conditionalization.

---

Add a 'koji-containerbuild flatpak-build' subcommand that triggers
building a flatpak rather than a Docker container. Most of the code
and options are shared with koji-containerbuild, but a --module
option is added, and since a new koji/osbs can be assumed:

 * The --git-branch option is required, as required by recent OSBS
 * source archive uploads are forbidden (did this ever work?),
   and the --no-progress option is removed.
 * The --release option is removed

In the builder plugin, the flatpak and module options are passed to osbs.

### DIFF
--- a/koji_containerbuild/cli.py
+++ b/koji_containerbuild/cli.py
@@ -61,7 +61,7 @@ def print_task_result(task_id, result, weburl):
     print_result(result)
 
 
-def parse_arguments(options, args):
+def parse_arguments(options, args, flatpak):
     def arches_parser(option, opt_str, value, parser):
         value = parser.values.ensure_value(option.dest, [])
         for arg in parser.rargs:
@@ -73,15 +73,22 @@ def parse_arguments(options, args):
         setattr(parser.values, option.dest, value)
 
     "Build a container"
-    usage = _("usage: %prog container-build [options] target <scm url or "
-              "archive path>")
+    if flatpak:
+        usage = _("usage: %prog flatpak-build [options] target <scm url>")
+    else:
+        usage = _("usage: %prog container-build [options] target <scm url or "
+                  "archive path>")
     usage += _("\n(Specify the --help global option for a list of other help "
                "options)")
     parser = OptionParser(usage=usage)
+    if flatpak:
+        parser.add_option("-m", "--module", metavar="NAME:STREAM[:VERSION]",
+                          help="module to build against")
     parser.add_option("--scratch", action="store_true",
                       help=_("Perform a scratch build"))
-    parser.add_option("--isolated", action="store_true",
-                      help=_("Perform an isolated build"))
+    if not flatpak:
+        parser.add_option("--isolated", action="store_true",
+                          help=_("Perform an isolated build"))
     parser.add_option("--arches", dest='arch_override',
                       action="callback", callback=arches_parser,
                       help=_("Requires --scratch. Limit a scratch build to "
@@ -109,10 +116,11 @@ def parse_arguments(options, args):
     parser.add_option("--channel-override",
                       help=_("Use a non-standard channel [default: %default]"),
                       default=DEFAULT_CHANNEL)
-    parser.add_option("--release",
-                      help=_("Set release value"))
-    parser.add_option("--koji-parent-build",
-                      help=_("Overwrite parent image with image from koji build"))
+    if not flatpak:
+        parser.add_option("--release",
+                          help=_("Set release value"))
+        parser.add_option("--koji-parent-build",
+                          help=_("Overwrite parent image with image from koji build"))
     build_opts, args = parser.parse_args(args)
     if len(args) != 2:
         parser.error(_("Exactly two arguments (a build target and a SCM URL) "
@@ -121,8 +129,18 @@ def parse_arguments(options, args):
     opts = {}
     if not build_opts.git_branch:
         parser.error(_("git-branch must be specified"))
-    for key in ('scratch', 'arch_override', 'epoch', 'yum_repourls',
-                'release', 'git_branch', 'isolated', 'koji_parent_build'):
+
+    keys = ('scratch', 'arch_override', 'epoch', 'yum_repourls',
+            'git_branch')
+
+    if flatpak:
+        if not build_opts.module:
+            parser.error(_("module must be specified"))
+        opts['flatpak'] = True
+        keys += ('module',)
+    else:
+        keys += ('release', 'isolated', 'koji_parent_build')
+    for key in keys:
         val = getattr(build_opts, key)
         if val is not None:
             opts[key] = val
@@ -131,8 +149,8 @@ def parse_arguments(options, args):
     return build_opts, args, opts, parser
 
 
-def handle_container_build(options, session, args):
-    build_opts, args, opts, parser = parse_arguments(options, args)
+def handle_build(options, session, args, flatpak):
+    build_opts, args, opts, parser = parse_arguments(options, args, flatpak)
 
     if build_opts.isolated and build_opts.scratch:
         parser.error(_("Build cannot be both isolated and scratch"))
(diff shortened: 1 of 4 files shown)


vrutkovs:
Looks good, but might need a rebase after #73 is merged

koji_containerbuild/cli.py
```diff
@@ -79,8 +85,9 @@ def handle_container_build(options, session, args):
     parser.add_option("--quiet", action="store_true",
                       help=_("Do not print the task information"),
                       default=options.quiet)
-    parser.add_option("--noprogress", action="store_true",
-                      help=_("Do not display progress of the upload"))
+    if not flatpak:
+        parser.add_option("--noprogress", action="store_true",
+                          help=_("Do not display progress of the upload"))
```
twaugh: Actually this doesn't work for container-build either.

koji_containerbuild/cli.py
```diff
@@ -135,6 +143,14 @@ def activate_session(session, options):
             parser.error(_("Destination tag %s is locked" % dest_tag['name']))
     source = args[1]
     opts = {}
+    if flatpak:
+        if not build_opts.module:
+            parser.error(_("module must be specified"))
+        if not build_opts.git_branch:
+            parser.error(_("git-branch must be specified"))
```
twaugh: Actually git-branch is required for container-build as well.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -40,6 +40,7 @@
 from osbs.api import OSBS
 from osbs.conf import Configuration
 from osbs.exceptions import OsbsValidationException
+from osbs.utils import split_module_spec
```
twaugh: Would be worth having this catch `ImportError` to retain compatibility with older versions of osbs-client.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -678,7 +687,23 @@ def handler(self, src, target, opts=None):
...
+        if flatpak:
+            module = opts.get('module', None)
+            if not module:
+                raise BuildError("Module must be specifed for a Flatpak build")
+
+            module_name, module_stream, module_version = split_module_spec(module)
+
+            data = {
+                'name': module_name,
+                'version': module_stream
```
twaugh: Nit: usually nice to use a trailing comma even for the last item, to make future diffs easier to read.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -687,11 +712,15 @@ def handler(self, src, target, opts=None):
...
 
         try:
-            auto_release = (data[LABEL_DATA_MAP['RELEASE']] ==
-                            LABEL_DEFAULT_VALUES['RELEASE'])
-            if auto_release:
-                # Do not expose default release value
-                del data[LABEL_DATA_MAP['RELEASE']]
+            # Flatpak builds append .<N> to the release generated from module version
+            if flatpak:
+                auto_release = True
```
twaugh: This skips the `del` a few lines down. Is that a problem?

koji_containerbuild/plugins/builder_containerbuild.py
owtaylor: Something like?

```
+try:
 +    from osbs.utils import split_module_spec
+    osbs_flatpak_support = True
+except ImportError, e:
+    osbs_flatpak_support = False

[...]

         if flatpak:
+            if not osbs_flatpak_support:
+                raise BuildError("osbs-client on koji builder doesn't have Flatpak support")
```

koji_containerbuild/plugins/builder_containerbuild.py
owtaylor: For the flatpak case the release field of data, if present, is not a meaningless object(), it's the release that we'll be appending .<N> onto. I suppose it could be considered confusing that it's not literally the release of the resulting artifact, but I think it's ifne.

koji_containerbuild/cli.py
owtaylor: I was leaving container-build untouched more because i didn't want to change it in an unrelated patch than because I thought that it didn't have the same problems. But certainly this would leave the code in a confusing state. I'll redo the branch with a commit to fix up container-build prior to the addition of flatpack support.

koji_containerbuild/plugins/builder_containerbuild.py
twaugh: Yes, like that.

koji_containerbuild/cli.py
```diff
@@ -96,15 +94,15 @@ def handle_container_build(options, session, args):
                       help=_("Git branch"))
     parser.add_option("--channel-override",
                       help=_("Use a non-standard channel [default: %default]"),
-                      default=DEFAULT_CHANNEL)
+                      default=DEFAULT_CHANNEL)<
```
twaugh: Typo in this commit?

koji_containerbuild/cli.py
```diff
@@ -94,11 +101,12 @@ def handle_container_build(options, session, args):
                       help=_("Git branch"))
     parser.add_option("--channel-override",
                       help=_("Use a non-standard channel [default: %default]"),
-                      default=DEFAULT_CHANNEL)<
-    parser.add_option("--release",
-                      help=_("Set release value"))
-    parser.add_option("--koji-parent-build",
-                      help=_("Overwrite parent image with image from koji build"))
+                      default=DEFAULT_CHANNEL)
```
twaugh: Think the 'cleanup' commit had a typo that was fixed in a later commit in this PR.

twaugh:
Merge conflicts.

Do the tests work for you? I get failures: https://gist.github.com/twaugh/aba3a559592059fa9266ac1e10b88f0e

owtaylor:
> Merge conflicts.
> 
> Do the tests work for you? I get failures: https://gist.github.com/twaugh/aba3a559592059fa9266ac1e10b88f0e

I'll fix up the merge conflicts and the (new since I wrote this patch) cli tests tomorrow. But if you are getting failures with my branch as non-rebased, I'm confused about why - it worked fine on my machine even with updated osbs-client. The gist isn't very helpful because of the typical flexmock-hides-actual-errors problem. You'll need to remove the `.times(0 if build_not_started else 1)` and then see what exception actually is.

koji_containerbuild/cli.py
owtaylor: Sorry, fixed in this rebase.

owtaylor:
OK, pushed the changes. Still don't know why it was failing for you - if you could retest/investigate that would be appreciated!

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -703,12 +715,30 @@ def handler(self, src, target, opts=None):
...
+
+            module_name, module_stream, module_version = split_module_spec(module)
 
+            data = {
+                'name': module_name,
+                'version': module_stream,
+            }
+
+            if module_version is not None:
+                data['release'] = module_version
```
twaugh: This block doesn't get covered during tests.

owtaylor: OK, should be fixed now with a bit of extra complexity to also test the cases when module is not specified or invalid.

lcarva:
Updated release notes draft:
> - Support for building flatpaks

PR #75, opened 2017-08-30
Cli arches and tests

mlangsdorf:
two commits to add support for the --arches argument and some tests of the CLI parser

    cli: write a test case for the arguments parser
    
    add a test case for the arguments parser. There are 9 arguments with
    binary or trinary values, so group the parametrization into two sets
    of 4 and 5 to prevent the combinations generating thousands of tests
    cases.

and

    cli: add --arches argument option
    
    --arches get passed on as arch_override

### DIFF
--- a/koji_containerbuild/cli.py
+++ b/koji_containerbuild/cli.py
@@ -62,6 +62,16 @@ def print_task_result(task_id, result, weburl):
 
 
 def parse_arguments(options, args):
+    def arches_parser(option, opt_str, value, parser):
+        value = parser.values.ensure_value(option.dest, [])
+        for arg in parser.rargs:
+            # stop when we hit a new arguments
+            if arg[:2] == "--" and len(arg) > 2:
+                break
+            value.append(arg)
+        del parser.rargs[:len(value)]
+        setattr(parser.values, option.dest, value)
+
     "Build a container"
     usage = _("usage: %prog container-build [options] target <scm url or "
               "archive path>")
@@ -72,6 +82,11 @@ def parse_arguments(options, args):
                       help=_("Perform a scratch build"))
     parser.add_option("--isolated", action="store_true",
                       help=_("Perform an isolated build"))
+    parser.add_option("--arches", dest='arch_override',
+                      action="callback", callback=arches_parser,
+                      help=_("Requires --scratch. Limit a scratch build to "
+                             "the specified arches. May specify multiple "
+                             "arches at once."))
     parser.add_option("--wait", action="store_true",
                       help=_("Wait on the build, even if running in the "
                              "background"))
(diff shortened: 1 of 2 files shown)


koji_containerbuild/cli.py
```diff
@@ -73,6 +72,9 @@ def handle_container_build(options, session, args):
                       help=_("Perform a scratch build"))
     parser.add_option("--isolated", action="store_true",
                       help=_("Perform an isolated build"))
+    parser.add_option("--arches", dest='arch_override',
```
lcarva: Should this be defined similarly to how `yum_repourls` is done? By using `action='append'`
If so, we'd need to change how arch_override is read. 
Not sure how the regular `build` subcommand implements this.

mlangsdorf: getArchList expects a list of strings. I think '--arches x86_64 powerpc64le' makes more sense than '--arches x86_64 --arches powerpc64le' but I'll adjust the code to pass a list like expected.

mlangsdorf: now supports both --arches arch1, arch2 and --arches arch1 --arches arch2.

Fixed.

lcarva: I think it might require: `--arches 'one two'` which seems to be very awkward.

lcarva: A custom parser seems overkill IMO...
Anyways, I'll approve but won't merge to give others a chance to take another look.

mlangsdorf: I'm pretty sure it works out to --arches arch1 arch2 but if I got it wrong, I can strip it back to --arches arch1 --arches arch2 which is very easy.

twaugh: The idea is that this works in the same way as it does for RPMs.
```
$ fedpkg build --help
[...]
  --arches [ARCHES [ARCHES ...]]
                        Build for specific arches
[...]
```

mlangsdorf: Fixed.

lcarva: I still see the custom parser. Is that intended?

From `pyrpkg/cli.py::cliClient::register_build_common` in rpkg, it looks like this would do it:
```
self.build_parser_common.add_argument(                      
    '--arches', nargs='*', help='Build for specific arches')
```

mlangsdorf: rkpg uses argparse, which supports parsing multiple arguments.
koji_containerbuild uses optparse, which does not: https://docs.python.org/2.6/library/optparse.html#callback-example-6-variable-arguments

lcarva: fair enough

lcarva:
Merge conflicts.

lcarva:
Updated release notes draft:
> - cli now takes --arches argument option

PR #76, opened 2017-09-06
builder_containerbuild: use get_orchestrator_build_logs api

jarodwilson:
Use the new get_orchestrator_build_logs API if available, to provide split
logs, one for orchestrator, one for each arch built.

Tested successfully in an osbs-box deployment.

Signed-off-by: Jarod Wilson <jarod@redhat.com>

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -335,13 +335,13 @@ class BuildContainerTask(BaseTaskHandler):
         log_filename = os.path.join(logs_dir, log_basename)
 
         self.logger.info("Will write follow log: %s", log_basename)
-        outfile = open(log_filename, 'w')
+        outfile = open(log_filename, 'wb')
         try:
             for line in log:
-                outfile.write("%s\n" % line)
-            outfile.flush()
+                outfile.write(("%s\n" % line).encode('utf-8'))
+                outfile.flush()
         except Exception, error:
-            msg = "Exception (%s) while reading build logs: %s" % (type(error),
+            msg = "Exception (%s) while writing build logs: %s" % (type(error),
                                                                    error)
             raise ContainerError(msg)
         finally:
@@ -356,15 +356,15 @@ class BuildContainerTask(BaseTaskHandler):
             if platform not in platform_logs:
                 prefix = 'orchestrator' if platform is None else platform
                 log_filename = os.path.join(logs_dir, "%s.log" % prefix)
-                platform_logs[platform] = open(log_filename, 'w')
+                platform_logs[platform] = open(log_filename, 'wb')
             try:
                 platform_logs[platform].write((entry.line + '\n').encode('utf-8'))
+                platform_logs[platform].flush()
             except Exception, error:
-                msg = "Exception (%s) while reading build logs: %s" % (type(error),
+                msg = "Exception (%s) while writing build logs: %s" % (type(error),
                                                                        error)
                 raise ContainerError(msg)
-        for platform, logfile in platform_logs.items():
-            logfile.flush()
+        for logfile in platform_logs.values():
             logfile.close()
             self.logger.info("%s written", logfile.name)
 
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -330,11 +330,16 @@ class BuildContainerTask(BaseTaskHandler):
         finally:
             watcher.clean()
 
-    def _write_combined_log(self, log, logs_dir):
+    def _write_combined_log(self, build_id, logs_dir):
         log_basename = 'openshift-incremental.log'
         log_filename = os.path.join(logs_dir, log_basename)
 
         self.logger.info("Will write follow log: %s", log_basename)
+        try:
+            log = self.osbs().get_build_logs(build_id, follow=True)
+        except Exception, error:
+            msg = "Exception while waiting for build logs: %s" % error
+            raise ContainerError(msg)
         outfile = open(log_filename, 'wb')
         try:
             for line in log:
@@ -348,8 +353,13 @@ class BuildContainerTask(BaseTaskHandler):
             outfile.close()
         self.logger.info("%s written", log_basename)
 
-    def _write_demultiplexed_logs(self, logs, logs_dir):
+    def _write_demultiplexed_logs(self, build_id, logs_dir):
         self.logger.info("Will write demuxed logs in: %s/", logs_dir)
+        try:
+            logs = self.osbs().get_orchestrator_build_logs(build_id, follow=True)
+        except Exception, error:
+            msg = "Exception while waiting for orchestrator build logs: %s" % error
+            raise ContainerError(msg)
         platform_logs = {}
         for entry in logs:
             platform = entry.platform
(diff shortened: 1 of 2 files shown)


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -330,27 +330,63 @@ def _incremental_upload_logs(self, child_pid):
...
-        except Exception, error:
-            msg = "Exception while waiting for build logs: %s" % error
-            raise ContainerError(msg)
         outfile = open(log_filename, 'w')
         try:
-            for line in build_logs:
+            for line in log:
                 outfile.write("%s\n" % line)
-                outfile.flush()
+            outfile.flush()
```
twaugh: Actually we want to flush after every line, otherwise when watching a stuck running build you don't know what's causing the delay because there are messages still buffered.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -330,27 +330,63 @@ def _incremental_upload_logs(self, child_pid):
...
+            platform = entry.platform
+            if platform not in platform_logs:
+                prefix = 'orchestrator' if platform is None else platform
+                log_filename = os.path.join(logs_dir, "%s.log" % prefix)
+                platform_logs[platform] = open(log_filename, 'w')
+            try:
+                platform_logs[platform].write((entry.line + '\n').encode('utf-8'))
+            except Exception, error:
+                msg = "Exception (%s) while reading build logs: %s" % (type(error),
+                                                                       error)
```
twaugh: While reading? Should that say "writing"?

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -330,27 +330,63 @@ def _incremental_upload_logs(self, child_pid):
...
+        self.logger.info("Will write demuxed logs in: %s/", logs_dir)
+        platform_logs = {}
+        for entry in logs:
+            platform = entry.platform
+            if platform not in platform_logs:
+                prefix = 'orchestrator' if platform is None else platform
+                log_filename = os.path.join(logs_dir, "%s.log" % prefix)
+                platform_logs[platform] = open(log_filename, 'w')
+            try:
+                platform_logs[platform].write((entry.line + '\n').encode('utf-8'))
```
twaugh: Can we call `.flush()` on that here, instead of at the end of the loop?

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -330,27 +330,63 @@ def _incremental_upload_logs(self, child_pid):
...
+
+    def _write_demultiplexed_logs(self, logs, logs_dir):
+        self.logger.info("Will write demuxed logs in: %s/", logs_dir)
+        platform_logs = {}
+        for entry in logs:
+            platform = entry.platform
+            if platform not in platform_logs:
+                prefix = 'orchestrator' if platform is None else platform
+                log_filename = os.path.join(logs_dir, "%s.log" % prefix)
+                platform_logs[platform] = open(log_filename, 'w')
```
twaugh: Is the mode OK? In atomic-reactor we needed `'wb'`: https://github.com/projectatomic/atomic-reactor/blob/1.6.26.2/atomic_reactor/plugins/exit_koji_import.py#L187

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -330,27 +330,63 @@ def _incremental_upload_logs(self, child_pid):
...
+            try:
+                platform_logs[platform].write((entry.line + '\n').encode('utf-8'))
+            except Exception, error:
+                msg = "Exception (%s) while reading build logs: %s" % (type(error),
+                                                                       error)
+                raise ContainerError(msg)
+        for platform, logfile in platform_logs.items():
+            logfile.flush()
+            logfile.close()
+            self.logger.info("%s written", logfile.name)
```
twaugh: I think we don't need this final loop. We'll perform `.flush()` after each `.write()`, and the `.close()` will happen automatically when `platform_logs` goes out of scope.

koji_containerbuild/plugins/builder_containerbuild.py
jarodwilson: Ah, probably. Copy-n-paste error mirroring the (probably also incorrect) code that was already there for writing out openshift-incremental.log.

koji_containerbuild/plugins/builder_containerbuild.py
jarodwilson: What about the self.logger.info() call? Both that and the .close() were intended to mirror what the prior code for the non-demuxed-logs case was doing.

koji_containerbuild/plugins/builder_containerbuild.py
jarodwilson: Yeah, it should probably be 'wb'. Not sure if we should do the same treatment, including .encode('utf-8') to the non-demuxed logging path as well, but kind of thinking maybe we should, so things stay consistent.

twaugh:
Oh, they were failing before this PR, for me at least. No new failures anyway.

jarodwilson:
So I'm at a stand-still on further improving test coverage, due to lack of knowledge about how flexmock and tmpdir work. Been trying to figure out how to get the mocked incremental_upload to dump out log files to verify their existence without any luck. We do verify both old and new logging code paths are taken, we just don't currently (and never did previously) have any validation that the expected log files exist in the end. It's been verified in osbx-box that we do, but not in automated testing. I'm somewhat inclined to ask that this be merged as-is right now though, and can continue working on additional testing coverage in an additional lower-priority PR, if necessary.

tests/test_kcb.py
```diff
@@ -220,14 +220,21 @@ def test_osbs_build(self, tmpdir, pkg_info, failure, orchestrator):
...
             .with_args(src['src'])
             .and_return(folders_info['dockerfile_path']))
         (flexmock(task)
             .should_receive('_write_incremental_logs'))
+        if orchestrator:
+            (flexmock(task)
+                .should_receive('_write_demultiplexed_logs'))
+        else:
+            (flexmock(task)
+                .should_receive('_write_combined_log'))
```
twaugh: All of these `should_receive()` calls *mask* the methods they name. Instead of calling those methods, flexmock uses a stub. As a result, we are not getting coverage for any of them.

Instead of doing that, I think we want to use flexmock to create a stub in the pretend OSBS client to provide a `get_orchestrator_build_logs` method, and similarly for `get_build_logs`.

lcarva: FYI, flexmock has the concept of spies to track when a method is called but without actually stubbing it out (`should_call` vs `should_receive`) https://has207.github.io/flexmock/user-guide.html#spies-proxies

jarodwilson: Yeah, I pretty much blindly copied the .should_receive('_write_incremental_logs') bit here, not realizing that these get stubbed out. I'm looking at the `get_orchestrator_build_logs` work done in osbs-client right now to try to get a better idea of what I need to do here.

lcarva:
Since adding unit test coverage to these changes has been proved to be problematic, let's have this merged and address unit tests for forked process at a later time.

lcarva:
Updated release notes draft:
> - `buildContainer` task splits orchestrator and workers log files incrementally

PR #77, opened 2017-09-07
 tests: throw OsbsOrchestratorNotEnabled if available

owtaylor:
With 5a3533a, if OsbsOrchestratorNotEnabled is available in osbs-client,
then the plugin looks specifically for this exception, and will not catch
OsbsValidationException; thus it's necessary to duplicate ObssOrchestratorNotEnabled
fallback logic in the test code.

### DIFF
--- a/tests/test_kcb.py
+++ b/tests/test_kcb.py
@@ -13,7 +13,10 @@ import os
 import os.path
 import koji
 from koji_containerbuild.plugins import builder_containerbuild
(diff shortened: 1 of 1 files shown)


owtaylor:
After some discussion on IRC, I realized that I attached only half the patch, and the commit message was wrong. Updated both.

twaugh:
Change is to test suite not main code so not release notes update needed.

PR #78, opened 2017-09-15
Use arch-override as str

lcarva:
build_containerbuild.py expects arch_override to be a string of
arches separated by spaces. This is exactly how the build command
works for building RPMs. To keep parity, a str is now sent.


lcarva:
Updated release notes draft:
> - cli now takes --arch-override argument option

PR #79, opened 2017-09-18
Move "isolated and scratch" check to parse_arguments to avoid crash in flatpak-build

vrutkovs:
Otherwise the build fails with:
```
AttributeError: Values instance has no attribute 'isolated'
```

/cc @owtaylor

### DIFF
--- a/koji_containerbuild/cli.py
+++ b/koji_containerbuild/cli.py
@@ -152,6 +152,9 @@ def parse_arguments(options, args, flatpak):
         opts['flatpak'] = True
         keys += ('module',)
     else:
+        if build_opts.isolated and build_opts.scratch:
+            parser.error(_("Build cannot be both isolated and scratch"))
+
         keys += ('release', 'isolated', 'koji_parent_build')
 
     if build_opts.arch_override:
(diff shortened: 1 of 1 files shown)


vrutkovs:
Unfortunately we can't - due to https://github.com/vrutkovs/koji-containerbuild/blob/master/koji_containerbuild/cli.py#L90 there is no `build_opts.isolated`

lcarva:
> Unfortunately we can't

I meant a few lines further down (~ line 142):

```python
    if flatpak:
        if not build_opts.module:
            parser.error(_("module must be specified"))
        opts['flatpak'] = True
        keys += ('module',)
    else:
        # Here
        if build_opts.isolated and build_opts.scratch:
            parser.error(_("Build cannot be both isolated and scratch"))
        keys += ('release', 'isolated', 'koji_parent_build')
```

vrutkovs:
Oh, I see. Sure, I'll rewrite that part

vrutkovs:
Updated

lcarva:
Not updating release notes draft because this is bugfix to an unreleased feature.

PR #80, opened 2017-09-25
Osbs client logs

lcarva:


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -293,15 +296,36 @@ def osbs(self):
                 build_conf = Configuration(conf_section='scratch')
             self._osbs = OSBS(os_conf, build_conf)
             assert self._osbs
+            self.setup_osbs_logging()
+
         return self._osbs
 
+    def setup_osbs_logging(self):
+        # Setting handler more than oce will cause duplicated log lines.
```
twaugh: Typo: oce instead of once

lcarva: Thanks! Fixed.

lcarva:
Updated release notes draft:
> - Builder plugin no longer loads `kojid` file as module
> - osbs-client logs are attached to `buildContainer` task

PR #81, opened 2017-09-25
Run koji-containerbuild tests in Travis

vrutkovs:
This PR is building a centos/fedora container and runs koji-containerbuild tests
and reports coverage data.

Note, that currently tests are py2 only, as there is no py3 koji version in repos 
yet.

In order to report results this repo should be added to Travis CI and Coveralls

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 # ## Linux
-.*
+.cache
+.coverage
+build/
 !.gitignore
 /dist/
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,7 @@
 # ## Linux
 .cache
 .coverage
-build/
-!.gitignore
+/build/
 /dist/
 
 # ## Python
--- a/koji-containerbuild.spec
+++ b/koji-containerbuild.spec
@@ -7,7 +7,7 @@
 
 Name:           koji-containerbuild
 Version:        0.7.4
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Koji support for building layered container images
 Group:          Applications/System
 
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -19,11 +19,7 @@ def run(self):
         'koji_containerbuild',
         'koji_containerbuild.plugins',
     ],
-    install_requires=[
-        "koji",
-        "osbs",
-        #"distutils",
-    ],
+    install_requires=[],
```
lcarva: Finally someone had the courage to remove these :)

lcarva:
"ImportError: No module named koji"

koji-containerbuild.spec
```diff
@@ -7,7 +7,7 @@
 
 Name:           koji-containerbuild
 Version:        0.7.4
-Release:        1%{?dist}
+Release:        2%{?dist}
```
lcarva: Why is this needed?

vrutkovs: Not sure if its required, as spec was changed, but no actual release was made. I'll keep this in case a new fix release will be done

.gitignore
```diff
@@ -1,5 +1,7 @@
 # ## Linux
-.*
+.cache
+.coverage
+build/
 !.gitignore
```
rcerven: remove this as well, now without '.*' it isn't useful

vrutkovs: Fixed

.gitignore
```diff
@@ -1,5 +1,7 @@
...
 /dist/
```
rcerven: this should be probably 'dist/'

lcarva: Slash in front of each item means only ignore starting at root of repo. It should probably stays with leading slash, and added to the other items too :)

koji-containerbuild.spec
rcerven: this is always 1 in atomic-reactor and osbs-client, do we really wanna change this?

lcarva:
Updated release notes draft:
> - Invalid dependencies removed from `setup.py`
> - Added `test/requirement.txt`

Not mentioning travis integration because that's irrelevant to released code.

PR #84, opened 2017-11-14
Disallow private branches for non-scratch container builds

bfontecc007:
This change will call callbacks with SCM info. If the build tag starts with rhel or RHEL, it will enforce scm policy. This has to be enabled by installing the scmpolicy plugin in koji build plugins. Under these conditions, branches starting with private are disallowed. 

Signed-off-by: Bret Fontecchio <bfontecc@redhat.com>

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -33,6 +33,11 @@ import dockerfile_parse
 import signal
 
 import koji
+
+# this is present because in some versions of koji, callback functions assume koji.plugin is
+# imported
+import koji.plugin
+
 from koji.daemon import SCM, incremental_upload
 from koji.tasks import ServerExit, BaseTaskHandler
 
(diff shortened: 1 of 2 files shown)


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -33,6 +33,7 @@
 import signal
 
 import koji
+import koji.plugin
```
lcarva: Let's add a comment explaining this import (https://pagure.io/koji/pull-request/693)

tests/test_kcb.py
```diff
@@ -283,6 +283,63 @@ def test_osbs_build(self, tmpdir, pkg_info, failure, orchestrator):
...
+        last_event_id = 456
+        koji_build_id = 999
+
+        options = flexmock(allowed_scms='pkgs.example.com:/*:no')
+
+        pkg_info = {'blocked': False}
+        session = flexmock()
+        (session
+            .should_receive('getLastEvent')
+            .and_return({'id': last_event_id}))
```
lcarva: Let's make sure we add `.once()` where necessary in these expectations.

bfontecc007: I added .times(2) to the callback check.

lcarva:
Updated release notes draft:
> - `preSCMCheckout` and `postSCMCheckout` callbacks are triggered

PR #85, opened 2017-11-20
Add compose_ids and signing_intent options

bfontecc007:
Allow --compose-id and --signing-intent parameters in koji-containerbuild cli. Pass compose_ids and signing_intent to osbs.

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -459,7 +459,8 @@ class BuildContainerTask(BaseTaskHandler):
     def runBuilds(self, src, target_info, arches, scratch=False, isolated=False,
                   yum_repourls=None, branch=None, push_url=None,
                   koji_parent_build=None, release=None,
-                  flatpak=False, module=None):
+                  flatpak=False, module=None, signing_intent=None,
+                  compose_ids=None):
 
         self.logger.debug("Spawning jobs for arches: %r" % (arches))
 
@@ -477,7 +478,9 @@ class BuildContainerTask(BaseTaskHandler):
             koji_parent_build=koji_parent_build,
             release=release,
             flatpak=flatpak,
-            module=module
+            module=module,
+            signing_intent=signing_intent,
+            compose_ids=compose_ids
         )
 
         results = [self.createContainer(**kwargs)]
@@ -488,7 +491,8 @@ class BuildContainerTask(BaseTaskHandler):
     def createContainer(self, src=None, target_info=None, arches=None,
                         scratch=None, isolated=None, yum_repourls=[],
                         branch=None, push_url=None, koji_parent_build=None,
-                        release=None, flatpak=False, module=None):
+                        release=None, flatpak=False, module=None, signing_intent=None,
+                        compose_ids=None):
         if not yum_repourls:
             yum_repourls = []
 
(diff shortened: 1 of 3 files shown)


tests/test_kcb.py
```diff
@@ -701,6 +724,7 @@ def test_cli_args(self, tmpdir, scratch, wait, quiet,
             assert build_opts.module == module
         else:
             assert build_opts.release == release
+        assert build_opts.compose_ids == compose_ids
```
lcarva: `assert build_opts.signing_intent == signing_intent` ?

tests/test_kcb.py
```diff
@@ -781,6 +805,57 @@ def test_isolated_scratch_restriction(self, tmpdir, scratch, isolated, valid):
         assert parsed_args == expected_args
         assert opts == expected_opts
 
+    @pytest.mark.parametrize(('additional_args', 'raises'), (
+        ({'compose_ids': None, 'signing_intent': None}, False),
```
lcarva: Could omit the keys that have `None` value.

tests/test_kcb.py
bfontecc007: added.

tests/test_kcb.py
bfontecc007: omitted.

lcarva: err.. I meant omit the key, not the scenario:
```
{},
{'compose_ids': [1, 2]},
...
```

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -506,6 +510,9 @@ def createContainer(self, src=None, target_info=None, arches=None,
         if not arches:
             raise ContainerError("arches aren't specified")
 
+        if signing_intent and compose_ids:
+            raise ContainerError("signing_intent used with compose_ids")
```
twaugh: This doesn't seem like the right exception to raise. Probably `koji.BuildError` is better for this. I see `ContainerError` is also raised if arches is not specified -- that also looks wrong to me.

tests/test_kcb.py
bfontecc007: oh I had no idea you could do that

koji_containerbuild/plugins/builder_containerbuild.py
lcarva: Good catch! It looks like this is also how it's done in RPM land:

```python
def getArchList(self, build_tag, h, extra=None):                                  
    # get list of arches to build for                                             
    buildconfig = self.session.getBuildConfig(build_tag, event=self.event_id)     
    arches = buildconfig['arches']                                                
    if not arches:                                                                
        #XXX - need to handle this better                                         
        raise koji.BuildError("No arches for tag %(name)s [%(id)s]" % buildconfig)
    tag_archlist = [koji.canonArch(a) for a in arches.split()]                    
    self.logger.debug('arches: %s' % arches)                                      
    if extra:                                                                     
        self.logger.debug('Got extra arches: %s' % extra)                         
        arches = "%s %s" % (arches,extra)                                         
    archlist = arches.split()                                                     
```
https://pagure.io/koji/blob/master/f/builder/kojid#_958

tests/test_kcb.py
```diff
@@ -781,6 +806,57 @@ def test_isolated_scratch_restriction(self, tmpdir, scratch, isolated, valid):
         assert parsed_args == expected_args
         assert opts == expected_opts
 
+    @pytest.mark.parametrize(('additional_args', 'raises'), (
+        ({}, False),
+        ({'compose_ids': [1, 2], 'signing_intent': None}, False),
```
lcarva: Let's also remove all keys that are set to `None`. In full this should be:
```python
({}, False),
({'compose_ids': [1, 2],}, False),
({'signing_intent': 'intent1'}, False),
({'compose_ids': [1, 2], 'signing_intent': 'intent1'}, True)
```

lcarva:
Updated release notes draft:
> - `compose_ids` and `signing_intent` task options

PR #86, opened 2017-11-27
Disallow use of yum_repourls with compose_ids

bfontecc007:
The parameters --repo-url and --compose-id cannot be used together.

Signed-off-by: Bret Fontecchio <bfontecc@redhat.com>

### DIFF
--- a/koji_containerbuild/cli.py
+++ b/koji_containerbuild/cli.py
@@ -130,7 +130,8 @@ def parse_arguments(options, args, flatpak):
                       help=_("Signing intent of the ODCS composes [default: %default]"),
                       default=None, dest='signing_intent')
     parser.add_option("--compose-id",
-                      help=_("ODCS composes used. May be used multiple times"),
+                      help=_("ODCS composes used. May be used multiple times. Cannot be"
+                             "used with signing-intent or repo-url"),
                       dest='compose_ids', action='append', metavar="COMPOSE_ID")
     if not flatpak:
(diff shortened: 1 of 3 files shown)


koji_containerbuild/cli.py
```diff
@@ -130,7 +130,8 @@ def parse_arguments(options, args, flatpak):
                       help=_("Signing intent of the ODCS composes [default: %default]"),
```
lcarva: Do we need to update the help text here too?

koji_containerbuild/cli.py
```diff
@@ -149,6 +151,9 @@ def parse_arguments(options, args, flatpak):
     if build_opts.signing_intent and build_opts.compose_ids:
         parser.error(_("--signing-intent cannot be used with --compose-id"))
 
+    if build_opts.compose_ids and build_opts.yum_repourls:
+        parser.error(_("--compose-id cannot be used with --repo-url"))
```
twaugh: This line doesn't get test coverage (actually neither does line 152).

bfontecc007:
I disabled testing in CentOS as a workaround for the bug in py.test we are experiencing.

koji_containerbuild/cli.py
bfontecc007: Added a unit test.

lcarva:
Not updating release notes draft because this is bugfix to an unreleased feature.

PR #87, opened 2017-11-30
tests: install specific versions of test package to avoid test packages update breaks

vrutkovs:
Signed-off-by: Vadim Rutkovsky <vrutkovs@redhat.com>

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -20,7 +20,6 @@ if [[ $OS == "fedora" ]]; then
   PKG_EXTRA="dnf-plugins-core git-core python$PYTHON_VERSION-koji"
   BUILDDEP="dnf builddep"
   PYTHON="python$PYTHON_VERSION"
-  PYTEST="py.test-$PYTHON_VERSION"
 else
   PIP_PKG="python-pip"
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -20,15 +20,13 @@ if [[ $OS == "fedora" ]]; then
...
   PYTHON="python$PYTHON_VERSION"
-  PYTEST="py.test-$PYTHON_VERSION"
 else
   PIP_PKG="python-pip"
   PIP="pip"
   PKG="yum"
   PKG_EXTRA="yum-utils git-core koji python-dockerfile-parse"
   BUILDDEP="yum-builddep"
   PYTHON="python"
-  PYTEST="py.test"
```
bfontecc007: Why these two deletions?

vrutkovs: There is no need to track which Py version are we using here, as there will always be just one pytest installed and it will be `/usr/bin/pytest` binary

bfontecc007: gotcha

lcarva:
Not updating release notes draft because this is a change solely to unit tests.

PR #88, opened 2017-12-12
travis.yml: F25 is EOL, use F27 instead

vrutkovs:


vrutkovs:
No need to updated release notes - test-only change

PR #89, opened 2018-01-10
Tests/disable rawhide

lcarva:


bfontecc007:
What's with the z?

lcarva:
> What's with the z?

As the commit message states, this is to allow users running test.sh locally when selinux is enabled.
From `man docker run`:

```
To change a label in the container context, you can add either of two suffixes :z or :Z to the volume mount.
These suffixes tell Docker to relabel file objects on the shared volumes. The z option tells Docker that two
containers share the volume content. As a result, Docker labels the content with a shared content label.
Shared volume labels allow all containers to read/write content.  The Z option tells Docker to label the
content with a private unshared label.  Only the current container can use a private volume.
```

lcarva:
Not updating release notes draft because this is a change solely to unit tests.

PR #91, opened 2018-01-16
Use with-open-as instead of open-try-finally-close

twaugh:


lcarva:
Updated release notes draft:
> - use with-open-as instead of open-try-finally-close

PR #94, opened 2018-04-04
test.sh: install PyYAML for osbs-client

twaugh:
We install osbs-client using --no-deps in order to avoid compiling and to catch missing dependencies for koji-containerbuild.

However, osbs-client also needs PyYAML, so install that RPM.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -37,7 +37,7 @@ fi
 if [[ $OS != "fedora" ]]; then $RUN $PKG install -y epel-release; fi
 $RUN $PKG install -y $PKG_EXTRA
 $RUN $BUILDDEP -y koji-containerbuild.spec
-# Install package
+# Install pip package
 $RUN $PKG install -y $PIP_PKG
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -37,16 +37,21 @@ fi
...
 
 # Install other dependencies for tests
+if [[ $PYTHON_VERSION == 3 ]]; then
+  OSBS_CLIENT_DEPS="python3-PyYAML"
+else
+  OSBS_CLIENT_DEPS="PyYAML"
+fi
+$RUN $PKG install -y $OSBS_CLIENT_DEPS
 
-# Install latest osbs-client by installing dependencies from the master branch
```
sosiouxme: This looks like an accidental deletion?

twaugh: Good catch, fixed.

twaugh:
Not updating release notes since this is just a change to the test code.

PR #103, opened 2018-08-27
No need to use psql when creating channel

tkopecek:
I've noticed this docs part. There is no need to interact with postgres via sql, as CLI provides --new option.


lcarva:
Not updating release notes draft because this is a change solely to docs.

PR #109, opened 2018-12-18
Add development requirements file

athos-ribeiro:
I would add a requirements.txt file instead, but here we are installing osbs-client from git master, so appending the -devel seems to be the right thing to do here.


athos-ribeiro:
No release notes needed. Development related change.

PR #111, opened 2019-01-11
copy instead of move metadata.json, becasue brew builder has source m…

rcerven:
…ounted as read-only


twaugh:
LGTM

PR #112, opened 2019-01-14
use incremental_upload for metadata.json, because source file is on

rcerven:
read-only filesystem

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -148,7 +148,7 @@ class FileWatcher(object):
             return
 
         for fname in results:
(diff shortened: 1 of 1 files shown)


MartinBasti:
I'm lost here, could you please explain me what happens there like for 10 years old?

rcerven:
we can't copy that file directly to work/task/number where it would appear in koji task output, because that filesystem is mounted read-only, and that metadata.json file is in work/koji-upload/..../metadata.json

so we are using same logic as is for logs, that is putting that file to local directory, and then let "_incremental_upload_logs" upload those files via koji upload method "incremental_upload"

rcerven:
rn updated: use incremental_upload for metadata.json, because source file is on read-only filesystem

PR #113, opened 2019-02-06
containerbuild: make sure the target exists before trying to tag it

mlangsdorf:
after retrieving target_info, check that it exists and raise an error if it doesn't instead of getting a confusing error when attempting to tag it.

Signed-off-by: Mark Langsdorf <mlangsdo@redhat.com>


mlangsdorf:
Added "confirm that a target exists before tagging it and raise an error message if it doesn't." to the release notes.

PR #115, opened 2019-03-08
allow yum_repo with compose_id

rcerven:
* OSBS-6662


rcerven:
rn updated: allow yum_repo with compose_id

PR #117, opened 2019-04-17
Add missing unit tests for build_containerbuild plugin

chmeliik:


tests/test_kcb.py
```diff
@@ -75,6 +75,117 @@ def test_get_repositories(self, repos):
             repositories.extend(repo)
         assert set(cct._get_repositories(response)) ^ set(repositories) == set([])
 
+    def _check_logfiles(self, log_entries, logs_dir, orchestrator):
```
MartinBasti: Could we have `check_orchestrator_logs` and `check_non_orchestrator_logs` as separated methods instead of closures? It may increase readability

chmeliik:
@MartinBasti updated PR

MartinBasti:
Tests change, no RN needed

PR #120, opened 2019-04-24
Remove dead code

twaugh:
This code, present since the original version of the cli_containerbuild module, has never been used.

Neither repo_id nor skip_tag have ever appeared elsewhere in the project. The condition can never have evaluated as truthy.


chmeliik:
LGTM

PR #122, opened 2019-06-07
enable building py3 package for rhel8

rcerven:
* OSBS-7497


MartinBasti:
Shouldn't be everything py2 related skipped for RHEL8 and Fedora 32?

MartinBasti:
We agreed on meeting that we should continue to ship py2 until py2 support is completely dropped

rcerven:
rn updated

PR #123, opened 2019-06-11
fix building when only py3 is available in rhel8

rcerven:

### DIFF
--- a/koji-containerbuild.spec
+++ b/koji-containerbuild.spec
@@ -36,12 +36,13 @@ Source0:        %{name}-%{version}.tar.gz
 BuildArch:      noarch
 BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 
-BuildRequires:  python
-BuildRequires:  python2-devel
 %if 0%{with python3}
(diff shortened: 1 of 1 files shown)


twaugh:
Do we want to build only the Python 2, and not Python 3, packages on RHEL-7? I thought we'd want both there.

koji-containerbuild.spec
```diff
@@ -36,12 +36,13 @@ Source0:        %{name}-%{version}.tar.gz
 BuildArch:      noarch
 BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 
-BuildRequires:  python
-BuildRequires:  python2-devel
 %if 0%{with python3}
 BuildRequires:  python3-devel
-%endif
+BuildRequires:  python3-setuptools
+%else
```
MartinBasti: This means we stop building both py2 and py3 packages for Fedora, I don't think that we agreed on dropping py2 packages from fedora

MartinBasti: Was agreed that we drop py2 from fedora

rcerven:
rn not updated, fix for unreleased change

PR #127, opened 2019-07-16
Add jsonschema validation for task options

chmeliik:
* OSBS-7672

The schema is used to validate arguments for the BuildContainerTask
handler() method, it validates the types of the individual arguments
(src: string, target: string, opts: dict or none) as well as the types
of parameters in the 'opts' dictionary. The schema for the 'opts' dict
is not complete, it allows unknown parameters because some options are
missing. This can be amended in the future.

### DIFF
--- a/koji-containerbuild.spec
+++ b/koji-containerbuild.spec
@@ -68,6 +68,13 @@ Requires:   koji-containerbuild
 Requires:   osbs-client
 Requires:   python-urlgrabber
 Requires:   python-dockerfile-parse
+%if 0%{with python3}
+Requires:   python3-setuptools
+Requires:   python3-jsonschema
+%else
+Requires:   python2-setuptools
+Requires:   python2-jsonschema
+%endif
 
 %description builder
 Builder plugin that extend Koji to communicate with OpenShift build system and
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -32,6 +32,9 @@ import traceback
 import dockerfile_parse
 import signal
 import shutil
+import json
+import jsonschema
+from pkg_resources import resource_stream
 from distutils.version import LooseVersion
 
 import koji
@@ -57,6 +60,8 @@ if LooseVersion(OSBS_VERSION) < LooseVersion(OSBS_FLATPAK_SUPPORT_VERSION):
     osbs_flatpak_support = False
 else:
     osbs_flatpak_support = True
+
+PARAMS_SCHEMA_RELPATH = '../schemas/build_params.json'
 
 
 # List of LABEL identifiers used within Koji. Values doesn't need to correspond
@@ -760,8 +765,13 @@ class BuildContainerTask(BaseTaskHandler):
         return (labels_wrapper.get_extra_data(), labels_wrapper.get_expected_nvr())
 
     def handler(self, src, target, opts=None):
-        if not opts:
+        with resource_stream(__name__, PARAMS_SCHEMA_RELPATH) as schema_file:
+            params_schema = json.load(schema_file)
+            jsonschema.validate([src, target, opts], params_schema)
+
+        if opts is None:
             opts = {}
+
         self.opts = opts
         data = {}
 
--- a/koji_containerbuild/plugins/hub_containerbuild.py
+++ b/koji_containerbuild/plugins/hub_containerbuild.py
@@ -48,8 +48,9 @@ def buildContainer(src, target, opts=None, priority=None, channel='container'):
 
     Returns the task ID
     """
-    if not opts:
+    if opts is None:
         opts = {}
+
     taskOpts = {}
     if priority:
         if priority < 0:
--- a/koji_containerbuild/schemas/build_params.json
+++ b/koji_containerbuild/schemas/build_params.json
@@ -1,0 +1,66 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "description": "Parameters for a koji containerBuild task.",
+
+  "type": "array",
+  "items": [
+    {
+      "type": "string",
+      "description": "Source URI."
+    },
+    {
+      "type": "string",
+      "description": "Build target."
+    },
+    {
+      "type": ["object", "null"],
+      "properties": {
+        "scratch": {
+          "type": "boolean",
+          "description": "Perform a scratch build?"
+        },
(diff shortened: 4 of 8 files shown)


chmeliik:
I have no idea why the fedora29+python2 combo fails, but it seems to fail when running `test.sh` from the `master` branch as well.

MartinBasti:
I'm curios if `builder_containerbuild.py` is better place for this check because `hub_containerbuild` just constructs a call which can be made without it anyway.

I'd place that check into `BuildContainerTask.handler`

MartinBasti:
For start we could check types for:
```
scratch=opts.get('scratch', False),
isolated=opts.get('isolated', False),
yum_repourls=opts.get('yum_repourls', None),
branch=opts.get('git_branch', None),
push_url=opts.get('push_url', None),
koji_parent_build=opts.get('koji_parent_build'),
release=release_overwrite,
flatpak=flatpak,
compose_ids=opts.get('compose_ids', None),
signing_intent=opts.get('signing_intent', None),
```

chmeliik:
Moved validation from hub_containerbuild to builder_containerbuild and added type checks for the values @MartinBasti suggested.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -58,6 +60,10 @@
 else:
     osbs_flatpak_support = True
 
+# root of koji_containerbuild module
+TOP_DIR = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+BUILD_OPTS_SCHEMA_PATH = os.path.join(TOP_DIR, 'schemas', 'build_opts.json')
```
athos-ribeiro: nitpick: should this file be packaged under a data directory instead?

chmeliik: yeah that seems like a better place for it, done

chmeliik:
Now packaging the jsonschema definition file under /usr/share/

koji-containerbuild.spec
```diff
@@ -54,6 +54,11 @@ Summary:    Hub plugin that extend Koji to build layered container images
 Group:      Applications/System
 Requires:   koji-containerbuild
 Requires:   koji-hub
+%if 0%{with python3}
```
rcerven: shouldn't this be under %package builder ?

chmeliik: yes it should, fixed :)

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -58,6 +60,8 @@
 else:
     osbs_flatpak_support = True
 
+BUILD_PARAMS_SCHEMA_PATH = '/usr/share/koji-containerbuild/schemas/build_params.json'
```
twaugh: Could use `pkg_resources.resource_stream` to get this path I think. See how atomic-reactor does it: https://github.com/containerbuildsystem/atomic-reactor/blob/1.6.41/atomic_reactor/util.py#L1269-L1274

chmeliik: If I'm understanding it correctly, `pkg_resources.resource_stream` works in atomic-reactor because the json files are packaged under the `atomic_reactor` module (probably somewhere in `/usr/lib*/python*/atomic_reactor/`). Is there a way to use it for files packaged under `/usr/share/`? Or should the schema file for koji-containerbuild be packaged in the same way that atomic-reactor does it?

twaugh: We can stick with /usr/share, no big deal.

athos-ribeiro: Nice! this seems to be the proper place for the file.

Should the python package also install it here? I mean, what happens if this package is installed with `pip install`? You can take a look at how we do it in atomic-reactor for reference (check the `setup.py` file).

chmeliik:
Moved the installation of the jsonschema file from `koji-containerbuild.spec` to `setup.py`, thanks @athos-ribeiro (also squashed the related commits into one)

schemas/build_params.json
```diff
@@ -0,0 +1,66 @@
...
+          "type": ["array", "null"],
+          "items": {
+            "type": "integer"
+          },
+          "description": "ODCS composes used."
+        },
+        "signing_intent": {
+          "type": ["string", "null"],
+          "description": "Signing intent of the ODCS composes."
+        }
```
MartinBasti: Don't we need here to allow extra properties?

chmeliik: I think extra properties are allowed by default (there is a test for it, and it passes)

koji_containerbuild/plugins/builder_containerbuild.py
MartinBasti: `/usr` part cannot be hardcoded, `pip install --user` or installation into virtualenv places `share` directory to somewhere else. This part mus be detected dynamically.

chmeliik:
After a decent amount of googling, I did not find any sane way to make `pkg_resources` tell you where it installed data_files. What we could do instead is install the json file under `/usr/lib/python*/koji_containerbuild/`, which is the same thing that we do in atomic-reactor. Will squash commit if approved.

athos-ribeiro:
> After a decent amount of googling, I did not find any sane way to make `pkg_resources` tell you where it installed data_files. What we could do instead is install the json file under `/usr/lib/python*/koji_containerbuild/`, which is the same thing that we do in atomic-reactor. Will squash commit if approved.
> 
> E: forgot to remove `%files` entry from specfile, fixed now

We could use fallbacks. We could, for example, look for the file under /usr/share, then look for it under the project path. This way we would both comply with [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) and support user installations.

chmeliik:
> We could use fallbacks. We could, for example, look for the file under /usr/share, then look for it under the project path. This way we would both comply with [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) and support user installations.

We could, but it would IMO be slightly more ugly than it is now, and keeping json files in python modules seems to be a fairly common practice (jsonschema itself does this, for example) so I wouldn't worry about it unless you think having the file in /usr/share is important.

chmeliik:
Squashed commits.

MartinBasti:
Release notes:
- input parameters are validated by schema and should provide nicer error reporting when parameters mismatch

PR #128, opened 2019-07-16
Install missing dependency in testing script

chmeliik:
For fedora-based tests, dockerfile-parse was only installed as a side
effect of installing osbs-client. With recent changes to the osbs-client
specfile, tests for python2 on fedora have started failing. Install
dockerfile-parse explicitly to prevent this.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -26,7 +26,7 @@ else
   PIP_PKG="python-pip"
   PIP="pip"
   PKG="yum"
-  PKG_EXTRA="yum-utils git-core koji koji-hub python-dockerfile-parse"
+  PKG_EXTRA="yum-utils git-core koji koji-hub"
   BUILDDEP="yum-builddep"
   PYTHON="python"
 fi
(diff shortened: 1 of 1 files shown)


chmeliik:
@athos-ribeiro Is there any easy way to install the package directly from the spec file? Or at least install all the `Requires` dependencies? That way we would not have to specify runtime dependencies as `BuildRequires` instead of `Requires` and the tests could verify that the spec file defines all the required dependencies.

Either way, good point about installing dockerfile-parse from github, updated PR.

test.sh
```diff
@@ -61,6 +61,16 @@ $RUN rm -rf /tmp/osbs-client
 $RUN git clone https://github.com/projectatomic/osbs-client /tmp/osbs-client
 $RUN $BUILDDEP -y /tmp/osbs-client/osbs-client.spec
 $RUN $PIP install --upgrade --no-deps --force-reinstall git+https://github.com/projectatomic/osbs-client
+
+# Install dockerfile-parse in the same way as osbs-client
+$RUN rm -rf /tmp/dockerfile-parse
+$RUN git clone https://github.com/containerbuildsystem/dockerfile-parse \
```
MartinBasti: could we set `--depth <depth>` here to 1?

MartinBasti: I'm also not a big fan of cloning that repo, do we really need it? Cannot we just use pip with dependencies?

MartinBasti:
@athos-ribeiro I don't think we want to test Builddeps here. Distro-specific tests should be separated IMHO.

test.sh
chmeliik: Ok, not cloning dfp. Note that this unfortunately does not work for osbs-client, it seems that some of the `BuildRequires` dependencies are actually required for building (other than `devel` and `setuptools`).

MartinBasti:
NO release notes needed, change in testing

PR #130, opened 2019-08-15
Declare python package runtime dependencies

athos-ribeiro:
* OSBS-7672

Signed-off-by: Athos Ribeiro <athos@redhat.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,16 @@ class TitoDist(sdist):
     def run(self):
         subprocess.call(["tito", "build", "--tgz", "-o", "."])
 
+
+def get_requirements(requirements_file='requirements.txt'):
+    with open(requirements_file) as f:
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -7,6 +7,16 @@ class TitoDist(sdist):
     def run(self):
         subprocess.call(["tito", "build", "--tgz", "-o", "."])
 
+
+def get_requirements(requirements_file='requirements.txt'):
+    with open(requirements_file) as f:
+        return [
+            line.rstrip().split('#')[0]
```
MartinBasti: Should be order like this?
```line.split('#')[0].rstrip()```

How about lines like (works with pypi):
```
   package 
    # test
```

MartinBasti: ok :D

PR #132, opened 2019-08-16
Put jsonschema definition directly in python file

chmeliik:
The package is installed in a specific way, which means the json file
will not be found in the expected location. To simplify things, let's
take advantage of the fact that jsonschema validation works with python
dictionaries and put the definition directly in builder_containerbuild.py.


athos-ribeiro:
Can you also include an OSBS comment in the commit?

chmeliik:
Removed `package_data` and included OSBS issue number, thanks @athos-ribeiro

chmeliik:
I believe the coverage decrease is due to a reduction in number of relevant lines (the entire `PARAMS_SCHEMA` definition counts as only one relevant line and some lines such as imports and opening files were removed)

athos-ribeiro:
No release notes added. There is already an entry for this feature

PR #134, opened 2019-08-21
check for required labels name & components only,

rcerven:
allow other labels to be defined from env,
check build existence only when explicitly defined nvr

* OSBS-7697

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -26,7 +26,6 @@ import os
 import os.path
 import sys
 import logging
-import imp
 import time
 import traceback
 import dockerfile_parse
@@ -67,9 +66,19 @@ LABELS = koji.Enum((
     'COMPONENT',
     'VERSION',
     'RELEASE',
-    'ARCHITECTURE',
+    'NAME',
 ))
 
+# list of labels which has to be explicitly defined with values
+LABELS_EXPLICIT_REQUIRED = koji.Enum((
+    'NAME',
+    'COMPONENT',
+))
+
+# list of labels which has to be defined but value may be set via env
+LABELS_ENV_REQUIRED = koji.Enum((
+    'VERSION',
+))
 
 # Mapping between LABEL identifiers within koji and actual LABEL identifiers
 # which can be found in Dockerfile. First value is preferred one, others are for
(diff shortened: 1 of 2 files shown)


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -263,10 +243,18 @@ def get_additional_tags(self):
...
+        # check required labels, have to be defined explicitly and not via env
+        for label_id in LABELS_EXPLICIT_REQUIRED:
             assert label_id in data
             if not data[label_id]:
                 missing_labels.append(label_id)
+
+        # check for all labels, unless required or default values provided,
+        # they should be at least defined (even via env)
+        for label_id in LABELS_ENV_REQUIRED:
+            assert label_id in data
```
athos-ribeiro: It would be nicer to raise an exception here (in an `if` block) instead of calling assert (when doing so, we could also remove the other assertion above).

rcerven: this is just really dummy assert because, in get_data_labels all LABELS which includes the ones from LABELS_EXPLICIT_REQUIRED, LABELS_ENV_REQUIRED will be defined

so I just can leave this here, or just remove that assert, I just wanted to keep it consisten

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -263,10 +243,18 @@ def get_additional_tags(self):
...
+            if data[label_id] is None:
```
twaugh: I think this is a bit subtle, and takes a bit of staring at to work out why it works.

* we set `data[label_id]` to `None` earlier on, in get_data_labels
* we do *not* have the parent environment available in order to correctly substitute environment variables in label values, so e.g. `LABEL foo="$BAR"` will end up with data['foo'] being the empty string
* it's the missing_labels variable, and not the assert, which is the actual check.

Would it be more readable if we made the following changes?:
* don't set explicit `None` values in get_data_labels, just leave them out of the dictionary if they aren't in the Dockerfile
* remove the two asserts here, because they are not the checks we care about
* replace the checks with:
  * `if not data.get(label_id):` (for `LABELS_EXPLICIT_REQUIRED`) and
  * `if label_id in data:` (for `LABELS_ENV_REQUIRED`)

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -214,18 +206,16 @@ def get_labels(self):
     def get_data_labels(self):
         """Subset of labels found in Dockerfile which we are interested in
 
-        returns dict with keys from LABELS and default values from
-        LABEL_DEFAULT_VALUES or actual values from Dockefile as mapped via
+        returns dict with keys from LABELS and values from Dockefile as mapped via
```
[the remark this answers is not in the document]
rcerven: ok

koji_containerbuild/plugins/builder_containerbuild.py
rcerven: chaned

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -214,18 +206,16 @@ def get_labels(self):
...
         """
 
         parsed_labels = self.get_labels()
         for label_id in LABELS:
             assert label_id in LABEL_NAME_MAP, ("Required LABEL doesn't map "
                                                 "to LABEL name in Dockerfile")
-            self._label_data.setdefault(label_id,
-                                        LABEL_DEFAULT_VALUES.get(label_id,
-                                                                 None))
+            self._label_data.setdefault(label_id, None)
```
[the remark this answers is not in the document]
rcerven: removed

rcerven:
coverage decreased because of removed code

rcerven:
rn updated: - check for required labels name & components only, allow other labels to be defined from env, check build existence only when explicitly defined nvr

PR #135, opened 2019-09-09
Skip build option, to update just buildconfig for autorebuilds

rcerven:
* OSBS-7711

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -393,17 +393,20 @@ class BuildContainerTask(BaseTaskHandler):
             os.makedirs(path)
         return path
 
-    def _incremental_upload_logs(self, child_pid):
+    def _incremental_upload_logs(self, child_pid=None):
         resultdir = self.resultdir()
         uploadpath = self.getUploadPath()
         watcher = FileWatcher(resultdir, logger=self.logger)
         finished = False
         try:
             while not finished:
-                time.sleep(1)
-                status = os.waitpid(child_pid, os.WNOHANG)
-                if status[0] != 0:
+                if child_pid is None:
                     finished = True
+                else:
+                    time.sleep(1)
+                    status = os.waitpid(child_pid, os.WNOHANG)
+                    if status[0] != 0:
+                        finished = True
 
                 for result in watcher.files_to_upload():
                     if result is False:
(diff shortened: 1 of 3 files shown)


koji_containerbuild/plugins/cli_containerbuild.py
```diff
@@ -105,6 +105,9 @@ def parse_arguments(options, args, flatpak):
                       help=_("ODCS composes used. May be used multiple times. Cannot be"
                              " used with --signing-intent"),
                       dest='compose_ids', action='append', metavar="COMPOSE_ID", type="int")
+    parser.add_option("--skip-build", action="store_true",
+                      help=_("Don't create build, but just modify settings for autorebuilds"))
```
MartinBasti: I'd omit part about autorebuilds here. It can be used for any settings right?

rcerven: I mean yes it can be used for updating any buildconfig, but unless it is autorebuild, updating buildconfig without build is basically meaningless

athos-ribeiro: This should be meaningful for the users, right?

Could we word it like "Skip build and update buildconfig. Use this option to update autorebuild settings"

rcerven: sure sounds good will change

rcerven: changed

rcerven:
I have to handle None request from osbs-client will have to update tests to cover that

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -622,11 +628,28 @@ def createContainer(self, src=None, target_info=None, arches=None,
...
 
+        if build_response is None:
+            self.logger.debug("Build was skipped")
+
+            osbs_logs_dir = self.resultdir()
+            koji.ensuredir(osbs_logs_dir)
+            pid = os.fork()
+            if pid:
+                try:
+                    self._incremental_upload_logs(pid)
```
athos-ribeiro: Do we need the fork here? Since we are not waiting for build logs to be written, could we use koji.daemon.incremental_upload directly, or refactor `_incremental_upload_logs` to have a child pid as an optional param (or to behave differently when pid==0)?

rcerven: refactored and got rid of fork

rcerven:
rn updated: Skip build option, to update just buildconfig for autorebuilds

PR #151, opened 2019-12-11
don't check for build nvr existance for autorebuilds which have 'trig…

rcerven:
…gered_after_koji_task',

as they might be using add_timestamp_to_release, so the nvr from
Dockerfile isn't the build nvr which will be created

* OSBS-8377

Maintainers will complete the following section:
- [x] "Signed-off-by:" line is present in each commit
- [x] JSON/YAML configuration changes are updated in the relevant schema
- [ ] Pull request includes link to an osbs-docs PR for user documentation updates

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -867,6 +867,12 @@ class BuildContainerTask(BaseContainerTask):
         if not SCM.is_scm_url(src):
             raise koji.BuildError('Invalid source specification: %s' % src)
 
+        # don't check build nvr for autorebuild as they might be using
+        # add_timestamp_to_release
+        triggered_after_koji_task = opts.get('triggered_after_koji_task', None)
+        if triggered_after_koji_task:
+            expected_nvr = None
+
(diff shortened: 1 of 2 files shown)


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -867,6 +867,12 @@ def handler(self, src, target, opts=None):
         if not SCM.is_scm_url(src):
             raise koji.BuildError('Invalid source specification: %s' % src)
 
+        # don't check build nvr for autorebuild as they might be using
+        # add_timestamp_to_release
+        triggered_after_koji_task = opts.get('triggered_after_koji_task', None)
+        if triggered_after_koji_task:
+            expected_nvr = None
```
athos-ribeiro: Although code coverage passes through here, is this being properly tested?

rcerven: added test

rcerven:
> @rcerven, should we document this somewhere later?

I don't think so, we are basically just loosening build existence check

rcerven:
rn updated: don't check for build nvr existance for autorebuilds

PR #152, opened 2019-12-13
use orchestrator.log name for log in task for source container

rcerven:
* OSBS-8378

Maintainers will complete the following section:
- [x] "Signed-off-by:" line is present in each commit
- [x] JSON/YAML configuration changes are updated in the relevant schema
- [ ] Pull request includes link to an osbs-docs PR for user documentation updates

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -279,6 +279,7 @@ class BaseContainerTask(BaseTaskHandler):
         self._osbs = None
         self.demux = None
         self._log_handler_added = False
+        self.incremental_log_basename = 'openshift-incremental.log'
 
     def osbs(self):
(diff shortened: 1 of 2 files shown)


MartinBasti:
How regular builds determine name `orchestrator.log`?

twaugh:
The usual path is through _write_demultiplexed_logs, where the log file to use is decide on based on fields in the log lines.

athos-ribeiro:
It would be nicer to have this responsibility in the BuildSourceContainerTask class instead of adding an attribute for that, if feasible. Not sure how the best way to approach this would look like though. Maybe the log filename could be passed in a class attr or _write_combied_log could be refactored

rcerven:
changed

rcerven:
rn updated: use orchestrator.log name for log in task for source container

PR #185, opened 2020-05-08
containerbuild: check availability of services before starting

mlangsdorf:
add a new function, check_servers_up, that takes a map of urls to
service names/flags and uses the cachet API to confirm that all the
services are available before attempting to submit a build to
the build system.

Signed-off-by: Mark Langsdorf <mlangsdo@redhat.com>

- [ ] Commit messages are descriptive enough
- [ ] JSON/YAML configuration changes are updated in the relevant schema
- [ ] Pull request has a link to an osbs-docs PR for user documentation updates

# TODO
- [ ] Parse flags from configuration JSON
- [ ] Test cases for failing services causing the build to pause
I'd like a quick review that this is the right approach


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -859,6 +862,31 @@ def checkLabels(self, src, build_tag, label_overwrites=None):
             return (data['COMPONENT'], labels_wrapper.get_expected_nvr())
         return (data['COMPONENT'], None)
 
+    def check_servers_up(self, flags=None):
```
lcarva, line 865: @mlangsdorf, did you have a chance to look into the checkHost function? AFAIU, it's a callback that the koji builder will automatically call when deciding to take a task. https://pagure.io/koji/blob/master/f/koji/daemon.py#_1268

The benefit of using this approach is that the task remains in the free state so it's clear that it's actually running.

This needs further investigation because I'm not sure how often `checkHost` is called, and how often it's called. Trying this out in osbs-box would be ideal.

tests/test_builder_containerbuild.py
```diff
@@ -54,6 +55,60 @@ def mock_incremental_upload(session, fname, fd, uploadpath, logger=None):
 
 
 class TestBuilder(object):
+    @pytest.mark.parametrize(('flags', 'valid'), [
+        ({'http://internal.status.redhat.com/': ['osbs-deployments-allowed']}, True),
```
MartinBasti, line 59: Can we use some generic domains?

mlangsdorf: I was hoping to be able actually test the call, not just mock it.

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -557,6 +561,45 @@ def sigint_handler(*args, **kwargs):
...
+                    self.logger.error(msg)
+                    raise koji.BuildError(msg)
+                else:
+                    status_comp = status_json["data"][0]
+                    if status_comp["status"] != 1:
+                        self.logger.info("{} unavailable, deferring".format(service))
+                        return False
+        return False
+
+    def checkHost(self, hostdata):
```
lcarva, line 592: Maybe this needs to take `*args, **kwargs` ?

PR #202, opened 2021-03-18
Drop importlib_metadata from requirements.txt as it causes dependency issues; install via test.sh

hroncok:
I'd update the imports in the code as well, but they are not here. I don't understand why was this dependency added in the first place, but the commit message mentions CentOS and Python 2, so maybe it fixes a transitive depndency on importlib_metadata there, yet this can be avoided on Python 3.8+

Relevant problem in Fedora: [nothing provides python3.9dist(importlib-metadata) < 3 needed by koji-containerbuild-0.11.0-2.fc35.noarch](https://bugzilla.redhat.com/show_bug.cgi?id=1940324)

- [n/a] Code coverage from testing does not decrease and new code is covered

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,2 @@
-importlib_metadata<3
 jsonschema
 six
--- a/test.sh
+++ b/test.sh
@@ -112,7 +112,13 @@ function setup_kojic() {
     # pyrsistent >= 0.17 no longer supports python 2
     # pyrsistent is a dependency of jsonschema
     $RUN "${PIP_INST[@]}" 'pyrsistent==0.16.*'
+
+    # setuptools_scm >= 6 no longer supports python 2
+    $RUN "${PIP_INST[@]}" 'setuptools_scm<6'
   fi
+
+  # Workaround problems with dependency hell for older Pythons
+  $RUN "${PIP_INST[@]}" -U 'importlib_metadata<3;python_version<"3.8"'
 
   # Setuptools install koji-c from source
   $RUN $PYTHON setup.py install


MartinBasti:
Thank you.

This change caused build failure on el7, see packit error. Would you give a try to remove that dependency completely?

MartinBasti:
also please add signoff to the commit msg

hroncok:
> This change caused build failure on el7, see packit error.

```
error in koji-containerbuild setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers
```

That is a very old setuptools we have there :(

> Would you give a try to remove that dependency completely?

Sure, let me remove it and see what happens.

> also please add signoff to the commit msg

What am I signing?

hroncok:
So, importlib_metadata requirement is **not** required on runtime but only during tests. There seem to be a dependency somewhere in the test stack that does not specify dependencies properly or that clashes with some other dependency.

Let's pin importlib_metadata in the **test** requires instead and see if all the failed CI jobs use that one.

hroncok:
OK, this works. I cannot sign the commits because I am using the github web UI. You can squash them when merging and sign them with your own signature. Consider my addition public domain.

MartinBasti (approved):

rcerven (approved):

hroncok:
Thanks @ben-alkov for squashing.

MartinBasti:
No release notes, minor change affecting only tests

hroncok:
Not entirely true. This removes a runtime dependency.

MartinBasti:
So this ins not true?
> So, importlib_metadata is not required on runtime but only during tests....

hroncok:
importlib_metadata is not **actually** required on runtime, but it was specified as such.

MartinBasti:
Do you agree with this?

Release notes:
>  runtime dependency importlib_metadata removed from install requirements

hroncok:
I agree. I'd even say: 

> Unused runtime dependency importlib_metadata removed from install requirements

PR #226, opened 2021-12-07
When pipeline run didn't succeeded and also didn't finish, cancel it

rcerven:
* CLOUDBLD-8220

- [ ] Commit messages are descriptive enough
- [ ] JSON/YAML configuration changes are updated in the relevant schema
- [ ] Pull request has a link to an osbs-docs PR for user documentation updates

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -537,6 +537,14 @@ class BaseContainerTask(BaseTaskHandler):
 
         elif not has_succeeded:
             error_message = self.osbs().get_build_error_message(build_id)
+
+            if self.osbs().build_not_finished(build_id):
+                try:
+                    self.osbs().cancel_build(build_id)
+                except Exception as ex:
+                    self.logger.error("Error during canceling pipeline run %s: %s",
+                                      build_id, repr(ex))
+
             if error_message:
                 raise ContainerError('Image build failed. %s. OSBS build id: %s' %
                                      (error_message, build_id))
--- a/tests/test_builder_containerbuild.py
+++ b/tests/test_builder_containerbuild.py
@@ -615,12 +615,13 @@ class TestBuilder(object):
                 'koji_builds': [koji_build_id]
             }
 
-    @pytest.mark.parametrize('reason, expected_exc_type', [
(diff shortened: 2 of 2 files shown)


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -537,6 +537,13 @@ def sigint_handler(*args, **kwargs):
 
         elif not has_succeeded:
             error_message = self.osbs().get_build_error_message(build_id)
+
+            if self.osbs().build_not_finished(build_id):
+                try:
+                    self.osbs().cancel_build(build_id)
+                except Exception:
+                    pass
```
mkosiarc, line 545: Should this be empty?

rcerven: well we won't see logs from koji-c anyway so it doesn't matter at all

ben-alkov: So if canceling the build raises an exception, we just ignore it and continue?
It looks to me like the exception is "handled" by falling through to `if error_message`. Do we have any guarantee that `get_build_error_message` will *never* come up empty...?

rcerven: 1st I am getting original error message

then IF pipeline run isn't finished I cancel pipeline run (and I don't care anymore about it, I want to propagate only the original message, not that "additional" cancel failed)

rcerven: if error_message is empty that is handled bellow, but how is that relevant?

ben-alkov: I'm concerned about what happens if canceling the build fails with an exception

rcerven: build fails with the original error

chmeliik: Could we log the exception anyway? Maybe we won't see it, but just out of principle.

rcerven: sure

chmeliik (approved):
LGTM :+1:

ben-alkov (approved):

rcerven:
no rn, fix for unreleased feature

PR #232, opened 2022-02-22
split build logs by platform

hjmodi:
* CLOUDBLD-8139

Signed-off-by: Harsh Modi <hmodi@redhat.com>

- n/a JSON/YAML configuration changes are updated in the relevant schema
- [x] Pull request has a link to an osbs-docs PR for user documentation updates

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -3,7 +3,7 @@
 Acts as a wrapper between Koji and OpenShift builsystem via osbs for building
 containers."""
 
-# Copyright (C) 2015  Red Hat, Inc.
+# Copyright (C) 2015-2022  Red Hat, Inc.
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -390,8 +390,9 @@ class BaseContainerTask(BaseTaskHandler):
         except koji.ActionNotAllowed:
             pass
 
-    def _write_logs(self, build_id, logs_dir):
-        log_filename = os.path.join(logs_dir, self.incremental_log_basename)
+    def _write_logs(self, build_id, logs_dir, platforms: list = None):
+        logfiles = {'noarch': open(os.path.join(logs_dir, self.incremental_log_basename),
+                                   'wb')}
         self.logger.info("Will write follow log: %s", self.incremental_log_basename)
         try:
             logs = self.osbs().get_build_logs(build_id, follow=True, wait=True)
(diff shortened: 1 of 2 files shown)


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -396,25 +397,44 @@ def _write_logs(self, build_id, logs_dir):
...
+                )
+            else:
+                task_platform = 'noarch'
 
-                if user_warnings.is_user_warning(line):
-                    user_warnings.store(line)
-                    continue
+            if task_platform not in logfiles:
+                log_filename = self.incremental_log_basename.replace('osbs-build',
+                                                                     f'osbs-build-{task_platform}')
```
chmeliik, line 423: For reference, this is how OSBS 1 handles platform logs: https://github.com/containerbuildsystem/koji-containerbuild/blob/c1201d7551483e7459c08fd1098ecb27e9f3ef77/koji_containerbuild/plugins/builder_containerbuild.py#L400

Notably, the log filenames for the per-arch logs should be just "{platform}.log" (like in atomic-reactor)

lkolacek (approved):

rcerven (approved):

koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -773,7 +792,8 @@ def createContainer(self, src=None, target_info=None, arches=None,
         except OsbsValidationException as exc:
             raise ContainerError('OSBS validation exception: {0}'.format(exc))
 
-        return self.handle_build_response(self.osbs().get_build_name(build_response))
+        return self.handle_build_response(self.osbs().get_build_name(build_response),
+                                          platforms=arches)
```
chmeliik, line 796: What is the relationship between the `arches` here and the platforms that actually end up being used in atomic-reactor (determined by the check_and_set_platforms plugin)?

If these arches are not guaranteed to be a superset of the platforms actually used, that would be a problem. For example, let's say the build will run s390x tasks, but because s390x was not in arches, koji-c will not check for that platform. That means s390x logs will go to the noarch log.

chmeliik: It kind of looks like there _is_ a nearly impossible corner case where this would be a problem. If someone uses the `arch_override` option and specifies a non-canonical name for one of the platforms supported by the build tag (e.g. [ia32e for x86_64](https://pagure.io/koji/blob/77aec93b98b6bfe2b870e41c5c28949d2ba06412/f/koji/__init__.py#_1213)), this non-canonical name would be passed to atomic-reactor.
https://github.com/containerbuildsystem/koji-containerbuild/blob/0667b30ff9cf63cc285d23905ee77b9c72016497/koji_containerbuild/plugins/builder_containerbuild.py#L785-L800

But that would cause far bigger issues than just misplacing the logs, plus arch_override is only allowed for scratch and isolated builds anyway.

In all cases except this (IMO irrelevant) one, it appears the arches _are_ a superset.

chmeliik (approved):

chmeliik:
Would it be possible to document how per-platform logs work in osbs-docs? It seems important to mention that taskrun names affect how their logs will be handled

+ it would be nice to also explain it a bit more in the commit message

koji_containerbuild/plugins/builder_containerbuild.py
hjmodi: yes, except for `arch_override`, `arches` are superset of `platforms` we get from the `check_and_set_platforms` plugin.

hjmodi:
PR for docs update https://github.com/containerbuildsystem/osbs-docs/pull/208

hjmodi:
updated commit message as well

PR #233, opened 2022-03-04
change default channel for buildSourceContainer to container-source

rcerven:


ben-alkov (approved):

chmeliik:
Could somebody remind me why we need it?

rcerven:
> Could somebody remind me why we need it?
> 
> Is it just to make sure we have the correct default, or does this fix some problems for integration tests or something?

if we have default channel container, we need koji rule, but when there is koji rule, we can't use koji channel override

chmeliik (commented):
Noticed that the docstring still says the default is "container"

Other than that LGTM

rcerven:
> Noticed that the docstring still says the default is "container"
> 
> Other than that LGTM

removed that 'default channel' from the comment as it duplicates explicit param ...

chmeliik (approved):
LGTM :+1:

PR #234, opened 2022-03-09
Do not try to get koji-build-id of scratch tasks

chmeliik:
CLOUDBLD-8889

Scratch tasks do not create Koji builds => there is no koji-build-id in
the build_results.

Fix unit test mocking accordingly.

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -588,7 +588,8 @@ class BaseContainerTask(BaseTaskHandler):
                          repositories)
 
         koji_build_id = None
-        if has_succeeded:
+        if has_succeeded and not self.opts.get('scratch'):
+            # Only successful, non-scratch tasks create Koji builds
             # For backward compatibility reasons, koji_build_id has to be a string
             koji_build_id = str(build_results['koji-build-id'])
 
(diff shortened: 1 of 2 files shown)


koji_containerbuild/plugins/builder_containerbuild.py
```diff
@@ -588,7 +588,8 @@ def sigint_handler(*args, **kwargs):
                          repositories)
 
         koji_build_id = None
-        if has_succeeded:
+        if has_succeeded and not self.opts.get('scratch'):
+            # Only successful, non-scratch tasks create Koji builds
             # For backward compatibility reasons, koji_build_id has to be a string
             koji_build_id = str(build_results['koji-build-id'])
```
chmeliik, line 594: Would you prefer to check `if 'koji-build-id' in build_results`? That would be safer, on the other hand the koji-build-id should be required for all successful, non-scratch task, so maybe failing with a KeyError when it is not there is what we want?

rcerven: yeah this looks good, for non-scratch it should be there otherwise users would complain

rcerven (approved):

ben-alkov (approved):

chmeliik:
No release notes, fix for unreleased feature

PR #262, opened 2023-01-20
builder: document compose_ids behavior

ktdreyer:
Update the JSON schema for `compose_ids` to explain that when a user specifies `compose_ids`, OSBS will not generate new composes.

### DIFF
--- a/koji_containerbuild/plugins/builder_containerbuild.py
+++ b/koji_containerbuild/plugins/builder_containerbuild.py
@@ -710,7 +710,13 @@ class BuildContainerTask(BaseContainerTask):
                         "items": {
                             "type": "integer"
                         },
-                        "description": "ODCS composes used."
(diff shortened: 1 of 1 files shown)


ktdreyer:
Related rpkg `--help` change: https://pagure.io/rpkg/pull-request/651

ktdreyer:
@rcerven How about this instead:

> A list of ODCS composes to use during the build. If you do not set this parameter, OSBS will request its own ODCS composes based on the "compose" settings in containers.yaml. If you set this parameter, OSBS will not request its own ODCS composes, and it will only use the exact ones you specify here.

ktdreyer:
Thanks, I've updated with this newer text.

rcerven (approved):

ktdreyer:
I've rebased on the latest `master` for #264

ben-alkov (approved):

